PR #71, opened 2026-06-24
fix(worktree): repoint decomp gfx-script shebangs to env python3

NicoviGH:
## Problem
A fresh worktree must **regenerate** the decomp graphics (the primary checkout already has them built, so it never hits this). But the decomp gfx scripts hard-code `#!/bin/python3`, which does not exist on macOS (`python3` is `/usr/bin/python3`). The Makefile runs them via their shebang, so a fresh worktree dies at the first graphics target:

```
scripts/gfxtools/tsa_generator.py: /bin/python3: bad interpreter: No such file or directory
make[1]: *** [graphics/efxdragon/DemonLightBg_Close_10.fetsa1.bin] Error 126
```

Hit this live while building the ch02 dialogue worktree (#70).

## Fix
`worktree-setup.sh` now repoints those shebangs to `#!/usr/bin/env python3` right after the submodule init, **guarded on `/bin/python3` being absent** so Linux/CI is untouched. The `fireemblem8u` submodule is a build artifact, so the edit is local-to-the-worktree and never committed.

## Validation
Mechanism verified by hand during the ch02 build: the fresh worktree failed pre-patch and went fully green (ROM + `verify_text` 0 runaway) after the exact same `sed`/`perl` repoint. Shell `bash -n` clean.

### DIFF
--- a/tools/worktree-setup.sh
+++ b/tools/worktree-setup.sh
@@ -105,6 +105,18 @@ fi
 echo ">> initialising the fireemblem8u submodule in the worktree (from local objects)"
 ( cd "$WT_PATH" && git submodule update --init fireemblem8u )
 
+# A fresh decomp checkout's graphics scripts hard-code `#!/bin/python3`, which doesn't exist
+# on macOS (python3 lives at /usr/bin/python3). The Makefile runs them via their shebang, so a
+# fresh worktree fails to REGENERATE graphics (the primary checkout already has them built, so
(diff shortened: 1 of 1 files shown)


NicoviGH:
Redundant — `tools/build.sh` already applies this exact shebang fix (lines 28-35), and per the design-placement rule ("don't scatter one decision across desks") it should live only there. Root cause was using a bare `make` instead of `tools/build.sh test`, which the HANDOFF explicitly warns against. Closing.

PR #115, opened 2026-07-02
feat(#11): cosmetic d20 nat-20 crit flourish

### LINKED ISSUE #11: Cosmetic d20 crit flourish
When an FE crit fires, play a brief 'd20 lands on 20' flourish in the battle animation. Cosmetic only — FE crit rate decides the crit (decisions.md).

NicoviGH:
Closes #11 (anim-mode scope). The game's thesis moment: when FE's own crit math fires, a gold d20 showing 20 pops over the battle.

**Preview (rendered from the exact injected bytes):** [docs/demo/d20-crit-flourish-preview.png](https://github.com/NicoviGH/manchego-stars/blob/feat/11-d20-crit/docs/demo/d20-crit-flourish-preview.png)

## How (decomp-traced)
- FE8 rules a crit in `banim-battleparse.c`; the C08 command fires the white crit flash (`banim-efxhit.c`) without blocking the anim script; the flash tears BG1 down after 17 frames.
- The hook (`engine_hooks._inject_crit_d20_flourish`) starts a cloned SpellFx proc **at that teardown** — sequenced BG1 ownership, so neither the flash nor combat pacing changes, and **FE crit math stays the sole trigger**. Centered HUD overlay via the non-mirrored tilemap copy (the "20" never mirrors with attacker side); ~46-frame hold, then standard teardown.
- **Engine/content split respected:** the hook is campaign-agnostic and guarded in `check_engine_guards_present`; the art is the campaign's (`battle_anims/d20-crit.png`). No asset → no flourish → pure vanilla.
- Asset pipeline: PNG → 4bpp sheet + palette + 30×20 TSA in **stored-form GBA LZ77** (literal-only blocks — always-valid `LZ77UnCompWram` input, no compressor vendored).

## Verification
- `test_crit_flourish.py` (5 tests): LZ roundtrip vs a reference decoder, bin shapes, TSA reference validity, and a **decode-back pixel-exact roundtrip** of the injected bytes vs the source art. Its >15-color test caught and fixed a crash in the error path.
- Full suite + drift guard green; CI compiles the patched `banim-efxhit.c` under agbcc.
- Deferred (in the ADR): map-battle (no-anim) crits — different rendering path; in-emulator motion review via `PT_CHAR=<id> recordanim` at the next capture session.

https://claude.ai/code/session_01HTEKqFqEvJKsz1t874oxcR

---
_Generated by [Claude Code](https://claude.ai/code/session_01HTEKqFqEvJKsz1t874oxcR)_


NicoviGH (commented):
`/code-review` (focused engine-C pass). 4 findings, all addressed in 1025887 + 96973d8:

1. **Pierce crits showed no d20** (Wyvern Knight's Pierce routes to a separate, byte-identical teardown that wasn't hooked). Both teardowns now hooked. **Silencer stays excluded deliberately** — it has its own distinctive Chill flourish, and no MVP cast member can Silencer; documented in the hook + ADR.
2. **The 46-frame proc could clobber successor effects** (its unconditional `SpellFx_ClearBG1()` could blank a brave-second-hit die or a magic counter's spell background mid-display, and a lethal-crit proc could outlive the scene). Structural fix: the die is now **proc-less** — registered once, then the vanilla effect lifecycle owns BG1 (successors draw over it; scene exit resets it). The entire clobber class is gone.
3. **Size validator didn't track the real constraint**: `gSpellAnimBgfx` holds 232 tiles (0x1D00 B); the old 240×120 dims check permitted overflow. Now capped by tile count with the buffer cited. (Current asset: 57 tiles.)
4. **`data_banim.s` append wasn't idempotent** (unlike the C-side `MS #11` guard) — a double-run in one process duplicated assembler labels. Now guarded symmetrically; verified with a double-injection probe.

Also verified locally end-to-end this round: patched `banim-efxhit.c` compiles under a locally-built agbcc, the full ROM links (`make` green, `verify_text` 3404 messages / 0 runaway), and the preview PNG decodes from the exact injected bytes.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #60, opened 2026-06-19
Playtest: greedy clear-bot — real-combat chapter completion (#49 platform, brick 2)


NicoviGH:
## Greedy clear-bot — real-combat chapter completion (#49 platform, brick 2)

Second brick of the #49 playtest platform, after the smoke liveness net (landed:
`scenarios.smoke` / `smoke_ch01`, PASS on prologue + ch01; `decisions.md` §Playtest platform
first brick). The smoke net proves a chapter **boots and runs crash/soft-lock-free**; this proves
**completability** — a chapter can actually be *won* — which the idle smoke net can't, and which is
the precursor to the LLM-player (a rule-based policy first, then swap it for an LLM).

**Why not just generalize `scenarios.win`:** it *cheats* — `pokeFrail` the boss to 1 HP and hardcodes
`CHAR_SEPHEK`. A genuine clear needs real combat + generic boss detection.

### Design decisions
- **Generic boss detection** (no hardcoded char ids): a red unit whose `CharacterData.attributes`
  (`pCharacterData` at unit `+0x00`, attributes at `+0x28`) has `CA_BOSS = (1 << 15)` set
  (`include/bmunit.h:326`).
- **Greedy real-combat strategy**, each player phase, per live non-exhausted unit: if it can reach a
  tile adjacent to an attackable enemy, attack the best target (prefer the boss when reachable, else the
  nearest enemy it can damage) via the existing `marchToward` / `moveUnit` / `chooseAttack`; otherwise
  advance toward the boss. Win = the chapter advances (DefeatBoss) or a Seize fires (objective-dependent).
- **Pure, testable decision core**: `pickTarget(reachableTiles, enemies, prefs) -> {target, tile}` as a
  pure Lua function (range-1 melee first), TDD'd standalone like `liveness.lua` — decision logic split
  from emulator driving.
- **Lord-survival guard**: if the must-survive lord dies → FAIL.
- **Verdict**: PASS = chapter won within the turn budget; FAIL = budget exhausted / lord died / no boss
  found.

### Scope (vertical slices)
- [ ] Generic boss-detection helper + assert it finds the prologue boss (no `CHAR_SEPHEK`).
- [ ] Pure `pickTarget` core + Lua unit tests (melee/range-1, boss-preference, nearest-attackable).
- [ ] `scenarios.clear` (prologue): real-combat clear with no `pokeFrail`; assert a win on a built ROM.
[...]

PR #116, opened 2026-07-02
feat(#60): clear-bot last-mile breach — pure pickMove core + claimed-tile march

NicoviGH:
Advances #60 (the "last-mile breach / unjam logic" item from the 2026-06-25 diagnosis).

## What
- **`clearbot.pickMove`** — the march decision is now a pure, unit-tested core (like `pickTarget`): field-first with a Manhattan tiebreak (keeps momentum on BFS-field plateaus), **claimed/occupied-tile avoidance** (two marchers stop fighting over one chokepoint tile), and the **chokepoint-jam fallback** — when no reachable tile improves the unit's boss distance, push at the nearest enemy cork instead of idling.
- **TDD'd against the exact reported jam geometry** (boss on the ch01 gate at (21,7), flanking walls, escort at (20,10), two units from (14,8), move 5): the simulated march reaches attack range within 6 turns, where the old decision stalled 3 turns at field-distance 8. 7 new Lua asserts (16 total), all in `make test`.
- **Harness wiring**: `clearUnitAct` consumes the core with a per-phase claimed set + other units' current tiles as blocked; a new diagnostic separates *decision* failures from *mechanical* move-commit failures (cursor/menu) — the distinction the previous stall logs couldn't see.

## Still open on #60 (needs an emulator session)
- Fair-play `clear_ch01` validation on a built ROM (this container has no mGBA).
- The thin 2-unit deploy: `reachCh01Map` mashes Fight! without Pick Units — the breach now has better decisions but still a thin field.

https://claude.ai/code/session_01HTEKqFqEvJKsz1t874oxcR

---
_Generated by [Claude Code](https://claude.ai/code/session_01HTEKqFqEvJKsz1t874oxcR)_

### DIFF
--- a/tools/playtest/clearbot.lua
+++ b/tools/playtest/clearbot.lua
@@ -55,6 +55,11 @@ end
 
 local function tileKey(x, y) return x .. "," .. y end
 
+-- Off-field sentinel: a tile the BFS never reached must score worse than ANY real
+-- field distance. Real distances are bounded by map area (FE8 maps are < 65536
+-- tiles); Manhattan by map perimeter -- both far under this.
+local OFFFIELD = 1000000
+
 -- pickMove(reachable, opts) -> {x,y} | nil -- the march decision when nothing is
 -- attackable this turn (#60 last-mile breach). Pure, like pickTarget: the scenario
 -- reads reachability/field/positions and owns the actual moving.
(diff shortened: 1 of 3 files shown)


NicoviGH (commented):
`/code-review` (focused Lua pass). 5 findings, all fixed in a506530 — including what looks like the actual root cause of the #60 jam:

1. **`selectAndReach(u)` used its 15×10 default window** inside `clearUnitAct`, clipping every ch01 reach list at x=14 — and the reported stall was at exactly (14,8), the clip boundary, at field-distance 8 = the BFS distance from the boundary to the boss. The march decisions never saw a tile past column 14 no matter how good they were. Bounds now thread through from `clearUntilAdvance`.
2. **Fallback oscillation**: the jam fallback targeted the globally-nearest enemy with no cap — a straggler behind the advance could pull a follower (whose best tile was merely claimed by its leader) up to a full move backward, every turn. Now only **cork** enemies (at least as boss-near as the unit, `fd(e) ≤ curFd`) trigger it, and among cork-near ties the boss-nearest tile wins.
3. **Non-discriminating test**: the old fallback test passed even with the fallback deleted (the cork sat on the goal axis). Replaced with an off-axis cork + a new straggler-behind test (17 asserts).
4. **nil-fd asymmetry**: a terrain-disconnected unit silently got the weakest (Manhattan) behavior; nil now counts as maximally far so the maximally-jammed case still gets the cork fallback.
5. **Magic 1000 sentinel**: undocumented overflow bound on serpentine maps; replaced with a documented `OFFFIELD` sized so on-field always beats off-field.

All Lua suites green. In-emulator `clear_ch01` validation still pending (no mGBA in this environment) — but finding 1 means the next emulator run tests a bot that can actually see the east half of the map.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #165, opened 2026-07-15
Marty battle animation and character-scoped green Flux


NicoviGH:
Add Marty’s three-pose Shaman battle animation using the established donor-prime pipeline. His Shillelagh remains mechanically Flux; only Marty’s Dark-magic battle effects receive a green palette treatment. The implementation must cover every Dark tome he can wield (not Flux alone), while normal Flux and all other units’ Dark effects remain unchanged. The engine path must be campaign-agnostic and data-driven.\n\nCloses the remaining Marty battle-art slice from #65.

Issue #65, opened 2026-06-23
Epic: Faked battle animations — static-frame custom anims for non-vanilla units (additive donor-prime)


NicoviGH:
The next visual-immersion lever after portraits/map-sprites: give **any unit that isn't a stock vanilla class** (custom PCs, named bosses, custom enemies) a *custom battle animation* — without hand-drawing ~50 motion frames. **Fake it with 1–3 static AI-generated frames + the engine's existing effects** (projectile, hit spark, crit flash, HP bar), reusing the lessons + tooling pattern from the portrait pipeline.

> **Brainstormed & direction locked 2026-06-22 (CLAUDE + Nicolas).** This issue IS the design artifact (we don't write spec docs). Settled decisions get a dated ADR in `docs/decisions.md` in the same commit as the first implementation slice (Milestone A) — same convention as #63.

## Definition of done
A unit with a `battle_anim:` block in its YAML fights on-screen with its custom art (e.g. an AI-generated RBG loosing an arrow), built **additively** so no vanilla class is consumed, authored as painlessly as a portrait. Validation milestone: **RBG end-to-end** (Milestone A).

## How FE8 battle anims work (decomp-grounded)
- One master table `banim_data[]` (`src/banim_data.c`); each class anim = `{abbr, modes, script, oam_r, oam_l, pal}`. **Self-sizing** (`banim_number = sizeof/…`) — no hardcoded cap; `animId` is `u16`/`int` in the battle path, so the table can grow.
- A class selects its entry via `ClassData.pBattleAnimDef` → `GetBattleAnimationId(unit, def, weapon)` → `animId`. FE8 also has a per-character unique-anim path (`GetBattleAnimationId_WithUnique`).
- Each entry has **13 modes** (attack_close/_critical/_range/_back, dodge_close/_range, stand/_close/_range, attack_miss). Each mode is a script of `banim_code_frame duration, sheet, frame, oam` + `banim_code_hit_normal` + `banim_code_call_spell_anim` (fires the projectile/spell). Frames are OAM object sets over a 4bpp (16-colour) tile sheet. Blobs are LZ77-compressed in ROM, rebuilt by the decomp's pattern rules on `make`.

## Locked architecture — additive "donor-prime" (NOT destructive reskin)
[...]

PR #167, opened 2026-07-15
fix(#65): register the spell-palette-tint hook (guard + PATCHED_DECOMP_FILES)

NicoviGH:
Two standards registrations the #165 green-Dark-magic hook (PR #166) shipped without.

## What was missing
1. **`check.py` guard** — `_patch_banim_spell_palette_tint` wasn't in `check_engine_guards_present`'s tuple, so a refactor could silently drop it (no gate would notice Marty's Flux reverting to the vanilla palette). Now guarded (DEFINED + CALLED).
2. **`PATCHED_DECOMP_FILES`** — `src/banim-efxmagic.c` and `src/banim-ekrutils.c` (the two TUs the hook patches) weren't registered for restore, so `restore_vanilla_sources` never reverts them. Same latent class as the earlier `icon.c` gap: **CI passes (fresh checkout) but a local rebuild in a dirty tree accumulates state.** (`ekrbattle.h` was already listed.) Both registered.

## Verification
- Restored both files to vanilla, rebuilt from scratch -> hook re-applies cleanly (`spell palette tints: 1 character/weapon row`), build green.
- `verify_text` 3404/0 · 95 unit tests green · `check.py` clean · no campaign id in the engine hook (boundary intact).

**No change to the feature** — only the two registrations we always require.

### DIFF
--- a/HANDOFF.md
+++ b/HANDOFF.md
@@ -5,18 +5,40 @@ live in `CLAUDE.md`; issue scope and backlog live in GitHub.
 
 ## Current state
 
-- Battle-animation review for RBG and Wolfram is complete in PR #161. Meesmickle's Shaman-based
-  animation is merged in PR #163 (`1cd65bd`).
-- The Tourmaline palette correction is separately merged as PR #162 (`f9ed1cc`); #161 is based on
-  that work and does not reintroduce the palette change.
-- Marty's approved Shaman battle animation and character-scoped green Dark-magic palette treatment
-  are complete in PR #166.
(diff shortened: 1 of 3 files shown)


NicoviGH:
Bundled a handoff refresh (`95bbc9c`, docs-only) that sequences the follow-up work for the next instance: merge this → **#168** (deeper spell-tint refactor) → resume **ch04/ch05** re-basing (#24/#25). Kept Codex's lean HANDOFF structure.

Issue #170, opened 2026-07-15
Parity engine: extract vanilla item economy (chests/villages/shops/drops) from decomp, not memory


NicoviGH:
## The gap

The parity engine (`difficulty.py`) models **enemy pressure** (rosters, levels, weapons) and **player throughput** — and does so accurately, straight from the decomp (`PARITY_REFERENCE_UDEFS` parses the real `events_udefs.c` arrays). But it has **no item-economy dimension at all**: no chests, villages/houses, shops, or drops; no gold values.

That blind spot is the actual red flag. Because the engine doesn't cover economy, the vanilla twins' payouts got **hand-transcribed from memory** into the ch04/ch05 brainstorm (issues #24/#25) — and were **wrong**:

| Twin | Brainstorm claimed | Decomp truth (verified @ HEAD) |
|---|---|---|
| FE8 Ch4 *Ancient Horrors* | "0 chests, one Iron Axe village, ~270g" | **4 chests, no villages**: Red Gem 5,000 + Iron Sword 460 + Javelin 400 + Iron Lance 360 = **~6,220g** |
| FE8 Ch5 *The Empire's Reach* | "~27,760g gifts, 0 chests" *(right total, but never machine-checked)* | **5 village/house gifts + Armory + Vendor, no chests**: Guiding Ring 10,000 + Dracoshield 8,000 + Secret Book 8,000 + Armorslayer 1,260 + Torch 500 |

The enemy side was never wrong — the engine had it. The economy side was wrong precisely because a human filled the gap the engine left.

## The fix — make economy a first-class, decomp-extracted parity dimension

Extend the parity tool to parse the vanilla twin's economy the same way it parses rosters (ground truth, never memory), and print it beside the enemy bar:

- **Chests** — `Chest(ITEM_x, ...)` entries in `src/events/chN-eventinfo.h`.
- **Village/house gifts** — `GIVEITEMTO` preceded by `SVAL(EVT_SLOT_3, <item id>)` in `chN-eventscript.h`.
- **Shops** — `Armory`/`Vendor` → `ShopList_*` arrays in `events_shoplist.c`.
- **Drops** — red units with the `.itemDrop` bit set (note the known Ch4 Entombed/Fetid-Claw case: *wields* it, drop bit unset → 0 drops).
- **Values** — `costPerUse × maxUses` from `data_items.c` (gems/boosters = full value; sell ≈ half).

[...]

PR #173, opened 2026-07-16
feat(#171): model recruit-flips + reinforcement timing in the parity engine

### LINKED ISSUE #171: Parity engine: model recruit-flips (convertibles) + reinforcement arrival timing, not just the turn-1 snapshot
## The gap

`difficulty.py` models a chapter as a **static starting snapshot**: every enemy is counted as present turn 1, at full threat + clear-load, on both the vanilla-parity side and ours. It has **no notion of two dynamics** that materially change real difficulty — and vanilla FE8 leans on both:

### 1. Recruit-flips (convertible enemies)
A unit you recruit mid-battle is scored as a **full enemy** and never credited as the ally it becomes. Verified on the ch04/ch05 twin: FE8 Ch5's `1× killing-edge Myrmidon L5` **is Joshua** — counted in the 23-enemy bar (threat 10.3 / clear-load 5.15) as if you fight him to the death, when in practice you flip him to your side. Our planned ch05 does the same beat (Sahnar, Joshua-donor, recruited via a Basil/Natasha escort), so the distortion is symmetric — but it means the *absolute* difficulty of both is overstated by ~one strong unit, and the ally-side benefit is invisible.

### 2. Reinforcement waves (arrival timing)
Vanilla Ch5 spawns bandit waves on **turns 2 / 6 / 8** (`EventListScr_Ch5_Turn`) that raid the villages. The parity registry folds those reinforcement arrays into the flat 23-count as if they're on the field at turn 1 — so a wave that's a *timed race* reads identically to a wave that's a *turn-1 alpha strike*. Clear-load is inflated at t1 and the turn-pressure curve is lost.

## Proposed direction

- **`convertible` tag** on an enemy entry (recruiter + method): scored at reduced enemy weight (you neutralize rather than grind it down), and its class/donor added to the **ally** projection for the post-flip turns. Makes the Joshua/Sahnar flip legible on both sides.
- **`arrives_turn` on enemy/reinforcement entries**: fold into clear-load (still must be cleared) but keep out of the turn-1 threat/slot figure; optionally surface a coarse pressure-over-turns note. Requires tagging the vanilla reinforcement arrays in `PARITY_REFERENCE_UDEFS` with their spawn turns (from `EventListScr_*_Turn`).
- Both stay a **static proxy** — the emulator playtest remains the arbiter — but they stop the proxy from lying about set-piece chapters (recruit-a-boss, defend-the-villages).

## Sibling
Same "the proxy misses a real dimension" family as **#170** (item economy). Together they'd let `make difficulty` read story/difficulty/economy — including recruit and reinforcement dynamics — from HEAD instead of a designer's memory.

_Surfaced 2026-07-15 during the ch04/ch05 re-basing design (Nicolas: "the engine needs updates to account for recruiting and reinforcements")._

NicoviGH:
Closes #171. The sibling to #170: gives the parity engine the two **battlefield dynamics** the static turn-1 bar can't see — a convertible enemy you recruit instead of kill, and reinforcements that arrive over turns.

> **Stacked on #172 (#170).** Merge #172 first; this branches off it. The clean #171-only diff is `difficulty.py` dynamics functions + `test_difficulty.py` `BattlefieldDynamics`.

### Why
The static bar counts all 23 of Ch5 as a turn-1 kill — but Joshua (the killing-edge Myrmidon) is a **recruit**, not a grind, and 6 of the 23 are **turn 2/6/8 bandit waves**. So the bar over-reads the alpha strike and the total kill-count. Our ch05 does the same beats (Sahnar flip, eruption waves), so both sides need to model it to compare fairly.

### What (all auto-detected from HEAD; symmetric with our YAML)
- **Convertibles** — `CHAR`-macro targets in the twin's eventinfo (Ch5 → `CHARACTER_JOSHUA`; Ch4 → none); `convertible:` on our `enemy_units`. Counted as **turn-1 threat** (dangerous until flipped) but **discounted ×0.5 in clear-load** (neutralize, don't grind), plus a reported **recruit swing** = post-flip ally throughput.
- **Reinforcements** — `TurnEventPlayer`-loaded arrays + their arrival turn (Ch5 → turn 2/6/8); `arrives_turn:` on our `enemy_units`. Kept in clear-load, **excluded from turn-1 threat**.
- New **`BATTLEFIELD DYNAMICS`** report section — **additive**. The static `ENEMY-PRESSURE` verdict and the curve gate are untouched, so already-locked chapters don't shift; this shows what the proxy misses beside it.

On **FE8 Ch5**: dynamic turn-1 threat **8.1 vs static 10.3** (6 of 23 arrive later; Joshua flips), clear-load ~unchanged (you still clear them, just later). Exactly the distortion #171 set out to expose.

### Verify
72 difficulty + 94 build_campaign tests green · `make check` drift clean · curve verdicts unchanged (additive) · `git diff --check` clean.

Also renames `PARITY_REFERENCE_ECON` → `PARITY_REFERENCE_STEM` (now shared by #170's economy and this — both read the twin's event data from HEAD).

### DIFF
--- a/tools/difficulty.py
+++ b/tools/difficulty.py
@@ -709,6 +709,7 @@ def report(campaign, ch):
 
     _print_pressure(_chapter_pressure(chap))
     _print_economy(chap)
+    _print_dynamics(chap)
 
 
 def _chapter_pressure(chap, band=0.25):
@@ -768,8 +769,9 @@ def _print_pressure(p):
 # chests + village/house gifts + shops, valued from data_items.c. Drops are a documented v1
 # gap (the curated Ch4/Ch5 twins carry none) -- see #170.
 
-# parity_reference -> the decomp chapter file stem whose event data carries its economy.
-PARITY_REFERENCE_ECON = {
+# parity_reference -> the decomp chapter file stem (shared by the economy #170 and the
+# battlefield-dynamics #171 extractors -- both read that chapter's event data from HEAD).
+PARITY_REFERENCE_STEM = {
     'FE8 Prologue': 'prologue', 'FE8 Ch1': 'ch1', 'FE8 Ch2': 'ch2', 'FE8 Ch3': 'ch3',
     'FE8 Ch4': 'ch4', 'FE8 Ch5': 'ch5', 'FE8 Ch6': 'ch6', 'FE8 Ch13': 'ch13a',
 }
@@ -849,7 +851,7 @@ def vanilla_economy(parity_ref):
     """The vanilla twin's item economy, read from HEAD: chests, gifts (villages/houses/clear
     rewards), and shops, each valued in buy-gold. None if the reference has no mapped economy
     source (#170). Delivery context (# of villages/houses) is kept for the vehicle story."""
-    stem = PARITY_REFERENCE_ECON.get(parity_ref)
+    stem = PARITY_REFERENCE_STEM.get(parity_ref)
     if stem is None:
         return None
     info = bc.vanilla_decomp_text('src/events/%s-eventinfo.h' % stem)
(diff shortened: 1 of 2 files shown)


NicoviGH:
**Known v1 scope (found while characterizing Ch4):** the reinforcement detector catches `TurnEventPlayer` waves (Ch5's turn 2/6/8) but NOT **area/zone-triggered** spawns. Vanilla Ch4 *Ancient Horrors* has ~7 such reinforcements (3 Bonewalkers turn-2, 4 Revenants on zone-entry) triggered outside the Turn list, so `vanilla_enemy_groups('FE8 Ch4')` currently reports reinf 0 (dynamic threat == static). Turn-based detection is correct; area-triggered is a documented follow-up (parse `AreaEvent`/`LOCA`-style triggers). Not blocking — the Ch5 set-piece this PR targets is fully turn-based.

Issue #90, opened 2026-06-26
Enemy battle-anim import pipeline (FEditor→decomp): ch1 fire imps + ch3 kobolds


NicoviGH:
Follow-up split off the #65 party-battle-anim work (kept separate to not derail the PC milestone).

**State today:** the ch01 grunts ("Foaming Mugs goblins") have a custom **map sprite** only — the vendored **Fire Imp** (Alexsplode, FE-Repo; soldier/fighter reskin, red imp under enemy palette). In *battle* they animate as **vanilla Soldier/Fighter**. No goblin battle anim exists in the repo yet (the Fire Imp is a map sprite; FE-Repo map sprites don't ship battle anims).

**Task:** hunt FE-Repo for a goblin/imp **battle animation** (axe + lance, to match the soldier/fighter grunts) and wire it.

**Why it's cheap / no new slots:** the goblins are generic enemy *classes*, so this is **class-level** (one shared anim per grunt type), and they **already have clone classes** (`CLASS_BLST_REGULAR_EMPTY` / `CLASS_BLST_LONG_EMPTY`) from the map-sprite reskin (#21). A vendored anim attaches as a class-level `.pBattleAnimDef` on those existing classes — no new class slots, and the per-character `_u25` path (used for PCs) does NOT apply (generic enemies have no unique character id).

**Contrast w/ PCs:** named units (PCs, named bosses) get per-character `_u25` anims (no slot); generic enemy *classes* stay class-bound. See decisions.md (character-unique vs class-clone ADR landing with the #65 M-B PC work).

NicoviGH:
## FE-Repo trawl — shortlist (mobile-week research, 2026-06-29)

**Headline finding:** there IS a `Fire Imp by Alexsplode` battle anim in FE-Repo — but it's **magic-only** (one `6. Magic/` folder, scripted as a fire-throwing caster). So the issue's "no goblin battle anim exists in the repo yet" framing is wrong; what's *missing* is a melee (axe + lance) goblin/imp anim to pair with the already-vendored Alexsplode Fire Imp **map sprite**. Confirmed via `CREDITS.txt`: "Fire Imp animations by Alexsplode. Scripting by Alexsplode. Map sprites and mug edited by Alexsplode."

No single FE-Repo pack covers both axe AND lance with a goblin/imp aesthetic — mix-and-match is unavoidable, and a palette pass at vendor-time will be needed for silhouette parity.

[...]

PR #190, opened 2026-07-18
feat(pinky): imported flier battle anim (swoop), bound per-character _u25

NicoviGH:
Pinky (the army's flier — **he/him**, RBG's homunculus son) gets a real N-frame battle animation: a hover-and-swoop that the faked 3-pose PC path can't carry. His six Gemini poses become a genuine dive-bomb.

## Pinky vs vanilla Pegasus Knight (in-engine)
Pinky vs vanilla

Pinky solo:
Pinky swoop

## What & how
- **Merges the two banim pipelines:** imports a REAL animation via `feditor_to_banim` (the #90 enemy path) but binds it **per-character via `_u25`** (not per-class). `build_unit_battle_anim` dispatches on a `battle_anim.import` block; both the import and faked paths return the identical `{sheets,pal,motion_s}` shape, so the binding is byte-for-byte the same. New `pegasus` `BANIM_DONORS` row.
- **New `tools/poses_to_feditor.py`:** hi-res poses → FEditor frames. It's the *inverse* of `descale` — instead of pinning the feet, each pose sits at its own canvas spot so the per-frame shift **becomes the motion**. Arc lives in `poses.yaml`, cadence in `Pinky.txt`.
- Lanceless body-slam dive (matches his map sprite); pink impact swirl.

## Tuning (all on the TESTCH `recordanim` capture)
Faces the foe · smallest unit on the roster (idle 27×31) · ear un-clipped (his tail was dragging the anchor into the arena ceiling) · impact lands **on** target · flyback bounces up into the hover instead of moon-walking the dive · lingers at apex + impact like vanilla.

## Verification
- 245 tests green (`test_build_campaign`, `test_feditor_to_banim`, `test_ref_to_battleframe`, `test_difficulty`)
- `make` green (TESTCH capture PASS, class 0x48) · `check.py` clean · `git diff --check` clean
- ADR in `decisions.md` (Art & Audio)


NicoviGH:
Re-rendered the review GIFs with `make_gif.py` (every captured frame, no windowing, 8 fps) instead of my hand-rolled 44-frame/12fps version — that was the jumpiness, same tooling the spell recordings use. Refresh the PR description GIFs to see the smoother motion.

NicoviGH:
Fixed the dodge bug you spotted: the dodge modes were reusing the apex/mid frames (which sit up at the attack-arc position), so on a counter Pinky teleported up and bounced there — the 'jaggedly comes back up and stays in a mid-jump frame'. Dodge is now a small hop near home. GIFs re-rendered; refresh to see it.

NicoviGH:
Dodge reworked — he now hops **up + back** (away from the foe) instead of lurching forward, using a dedicated frame (the jump art placed behind home) and vanilla's dodge cadence.

Focused dodge clip so you can judge the timing in motion:
Pinky dodge

From the static frames the hop still lands a hair after the lance rather than exactly as it comes in — the exact timing is engine-synced to `C0E`, so tell me if it reads late and I'll pull it earlier (or snappier). Direction + "no more teleport-to-apex" should both be fixed.

NicoviGH:
Fresh-named clip so you're definitely not seeing a cached one (renamed to `-v2`; the stale cache on the same filename was likely why it kept looking unchanged — sorry, that wasted your time):

Pinky dodge v2

Honest status on the dodge: in the ROM data it's now the **mirror of the launch jump** — same height, hopping *back* (+18) instead of forward (−6). But I'll be straight: it's been a stubborn mode to nail because (a) the jump art is a *leap*, so it reads as "up" more than "a step back," and (b) the dodge timing is engine-synced and lands a beat late here.

Since the dodge only ever plays when an enemy attacks Pinky **and misses** (rare), I don't want to keep burning builds on it if it's good enough. Two options — your call:
1. **Good enough** — ship it, move on (attack swoop is the thing players see 99% of the time).
2. **Keep going** — I dig into how vanilla handles the backward dodge translation (its `dodge_to_back`) so the step-back reads cleanly, and pull the timing earlier.

NicoviGH:
You're right — the trigger's in the code, not on screen. C01 in the dodge script is `wait_hp_deplete` (`0x85000001`), the engine's hit-resolution sync (the same beat the MISS fires). I'd been treating it as a NOP and putting the hop *before* it, so Pinky landed before the miss resolved. Vanilla pekf holds its dodge pose *across* `wait_hp_deplete`; now Pinky does too, so he's airborne through the resolution instead of already back down.

Pinky dodge v3

NicoviGH:
Fresh full-anim GIF (current build, cache-busting name):

Pinky battle anim v3

NicoviGH:
Real fix this time, and an honest accounting.

**The fix:** `wait_hp_deplete` *pauses* the anim until the hit resolves. Vanilla (and all my earlier tries) put the hop *before* it, so the back-hop fired early at `start_dodge`. Putting the hop *after* `wait_hp_deplete` syncs it to the resolution — the dodge moved from f38 to f42, a hair before the lance-strike (f46), which is what you asked for.

**My screw-up:** every "dodge clip" I sent before was frames 44–74 — that's Pinky's **second attack** (a double), not the dodge (which is ~f42). So when you said "still hops forward," you were seeing the attack I'd mislabeled. Here's the *actual* dodge window:

Pinky real dodge

The dodge is a quick, rare beat (only when an enemy hits Pinky and misses). If the timing reads right to you now, I'd call it done. If it's still off, tell me relative to *this* clip and I'll nudge — but I won't keep guessing on the wrong frames again.

Issue #232, opened 2026-08-06
ch01 parks in the Beat-1 cutscene (last of the 7 gate failures)


NicoviGH:
Surfaced by the first live run of the #231 playtest matrix (`make matrix SUITE=gate`) on `main`'s content at `e0b61bd`. **These are pre-existing on `main`** — verified by re-running `ch04snag` through `main`'s own `run.sh` against the identical `CH04BOOT=1` ROM: byte-identical failure. Nothing in #231 touches `harness.lua` or `controller.lua`.

The gate suite is 14 scenarios, 3 ROM builds, ~5 minutes wall.

## PASS (7)

`controller_turn` · `gameover` · `retreat` · `recordunitlist` · `ch04moose` · `ch04village` · `ch04cottage`

### Cluster A — the #220 controller's `confirm_target` postcondition (5 scenarios)

`win` · `ch01` · `ch01win` · `lordfloor` · `ch04packmath`

Every one dies in the same place: the controller presses `A` on a chosen target, expects *"target selection starts the battle proc"*, and the observed `after` state is `player_map_idle` with the attacker already `unit_select_kind: turn_ended`.

```
"expected":"target selection starts the battle proc","input":"A",
"legal":["confirm_target"],"result":"fail:postcondition","state":"target_selection"
```

So the attack **did** resolve — the unit acted and its turn ended — but the postcondition never observed a battle proc. Two readings worth testing before fixing:

1. the postcondition observes too late (the battle proc has already come and gone between polls), or
2. the postcondition assumes an animated battle and no battle proc exists on the path actually taken.

`win`'s dump is the useful one: it caught `ekrBattleEnding` live with `battleTarget hp=0` and `gBanimDoneFlag [0]=1 [1]=1` — the boss was already dead while the check was still waiting, which points at (1).

`lordfloor` fails as *"never won ch00"* — same root cause one level up, since `winCh00` drives through `chooseAttack`.

Note the blast radius: this is the shared attack path, so it takes out the ch00/ch01 spine **and** the ch04 balance gate at once. It will do the same to ch05.

### Cluster B — two ch04 content regressions (2 scenarios)

- **`ch04snag`** — *"the snag broke but (4,9) is terrain 0x10, not a crossing (0x13) — the MapChange did not apply"*. The snag itself breaks (its own tile reaches `0x33`); the paired MapChange that turns the water into a crossing does not land. Shipped and gated in #215.
[...]

Issue #222, opened 2026-08-03
[Epic] Accelerate development with semantic playtest tooling


NicoviGH:
## Outcome

Reduce the elapsed time, manual commands, and diagnosis effort between a campaign/engine change and a trustworthy playtest answer.

This epic builds on the state-driven FE8 controller delivered by #220 / PR #221. That controller remains the shared UI-mechanics layer; these workstreams improve orchestration, observability, authoring speed, and pre-build validation around it.

## Success metrics

Track these before and after the epic:

- **One command** runs the requested live regression matrix.
- Each required ROM configuration is built **at most once per matrix run**.
- Normal output is a compact verdict/duration table; full traces and screenshots remain artifacts.
- A failed run identifies the **first semantic state/transition divergence**, not merely the final timeout.
- A typical new semantic scenario can be expressed in a small goal/assertion manifest without adding controller mechanics.
- Deterministically invalid chapter/event data fails before the expensive ROM build or emulator run.
- Record wall-clock duration, duplicate-build count, scenario commands, and time-to-first-actionable-failure so velocity gains are measured rather than assumed.

### 1. Playtest matrix runner — first priority

Create one orchestrator for grouped live verification.

Scope:

- Define a manifest mapping scenarios to ROM configurations such as canonical, `CH03BOOT=1`, and `CH04BOOT=1`.
- Build each configuration once, then run all scenarios assigned to it.
- Keep mGBA scenarios sequential where shared emulator/save state makes parallel execution unsafe.
- Emit a concise table containing build variant, scenario, verdict, duration, and artifact path.
- Preserve complete structured traces on disk; print only the failure summary and relevant tail by default.
- Support a fast changed-area subset and the full merge gate.
- Exit nonzero if any build or scenario fails.

Definition of Done:

- The #220 live matrix runs from one command.
- No ROM configuration is rebuilt unnecessarily within that run.
- Success output is short enough to scan immediately.
- Failure output names the failing semantic transition and links its artifacts.
- The matrix command is documented and covered by unit tests for grouping, ordering, and verdict aggregation.

### 2. State inspector and first-divergence report

Turn controller observations into a developer debugging surface.

Scope:

- Provide a one-shot human-readable snapshot and the underlying structured JSON.
- Summarize live state, chapter/turn/faction, cursor unit/selectability, active semantic menu commands, targets, dialogue wait, Preparations state, and relevant Proc ownership.
[...]

Issue #220, opened 2026-08-03
Playtest: state-driven FE8 controller — observe, enumerate, act, verify


NicoviGH:
## Goal

Replace timing-driven and guessed-input playtest automation with a reusable FE8 state-driven controller. A scenario must observe the live game state, enumerate legal actions, execute one justified input, and verify its postcondition before continuing.

This is a prerequisite for #24 / PR #219: the Chapter 4 moose and wolf-parley recorders exposed that the shared harness can enter Check Map/View Map, choose Rescue while intending Wait, and advance screens on blind cadences.

Blocks #219 until the shared controller lands and the Chapter 4 branch consumes it. Tracked under #49 Pipeline.

## Root cause

The current harness describes itself as closed-loop, but common paths still violate that contract:

- `bootToMap()` alternates A and START every 26 frames without identifying the active screen.
- `driveThroughPrep()` presses B from the Preparations main menu; FE8 maps that input to Check Map, so the harness visibly enters View Map before attempting Fight.
- unit and map menus sometimes assume row positions (for example, wrapping to the last row for Wait or End Phase);
- dialogue/cutscene loops sometimes press A on cadence instead of only while FE8 is waiting for dialogue input;
- recovery code tries additional buttons after an unexpected state, which can turn a timing miss into a valid but wrong action.

The immediate incident was `recordch04moose` selecting Rescue while trying to Wait. The broader defect is that the driver acts before it has authoritative evidence of state and legality.

## Locked architecture

Use the emulator/decomp boundary we already own. Do not use pixels, OCR, fixed window positions, or wall-clock sleeps as game-state truth.

The controller loop is:

`observe FE8 memory -> classify state -> enumerate legal actions -> execute one guarded input -> verify postcondition -> trace transition`

### 1. Observer

Read exact proc scripts, current proc callbacks, menus, and FE8 data structures to produce a named state. Initial coverage:

- Health & Safety input wait
- title-screen idle
- save/New Game menu input
- difficulty menu input
- chapter-intro input
- standard-event dialogue wait
- Preparations main menu
- Preparations View Map/map menu
- Pick Units
- player-phase map idle
- unit selected/movement range
- unit command menu
- map command menu
- battle/event transitions

[...]

PR #235, opened 2026-08-06
Repair the playtest waits the #220 rewrite left broken: gate 7/14 -> 13/14 (#232)

NicoviGH:
Fixes six of the seven gate failures in #232. **Gate: 7/14 → 13/14.**

> Follows #233 (now merged), which is what makes `make matrix` available to verify this. Rebased onto `main`. Supersedes the auto-closed #234, which GitHub closed when its base branch was deleted on merge.

## What was actually wrong

Not one defect — **six**, sharing one story. #220 correctly replaced blind cadence input with observed, guarded input. But `waitFor` had been mashing A every 50 frames, and that cadence was **load-bearing**: several waits were sized or written around it. When it went, they broke — and nobody noticed, because the ch00/ch01 scenarios aren't in HANDOFF's hand-run gate list and there is no CI for in-emulator playtests.

Every fix below is backed by a measurement, not a guess. Two of my early hypotheses were disproved by the evidence and are not in this PR.

| # | Root cause | Evidence that found it |
|---|---|---|
| 1 | `chooseAttack`'s combat budget was **1200** frames; an *unskipped* ch00 boss animation is **1238**. The old A cadence advanced the in-battle quote and pulled it under the wire. | Probe: battle proc alive at elapsed 0, gone at elapsed 1238 |
| 2 | `winCh00` passed no `stopWhen`. Killing Sephek fires DefeatBoss immediately, the chapter tears the map down, and the actor never becomes `US_UNSELECTABLE`. | A 9000-frame run ended at the **save menu** — win achieved — with the wait still spinning |
| 3 | `confirm_target` proved commit via `procs.battle` = `gProc_ekrBattle`, the battle *animation* proc. Every `pokeFastConfig()` scenario runs map combat, where it never spawns. | All three failures call `pokeFastConfig()`; trace `after` shows the attacker already `turn_ended` |
| 4 | `awaitControllerState` treated `dialogue_wait` as a fault and burned a flat 300-frame budget that no cutscene fits. | Stalls were `std_event` at chapter 2 turn 1 — the ch01 opening — on chapters the scenario went on to clear |
| 5 | The between-chapters save prompt is a **second** proc script (`gProcScr_SaveMenuPostChapter`) sharing `SaveMenu_SaveSlotSelectLoop`; the observer watched only `ProcScr_SaveMenu`. | Screenshot of the run parked on the save screen |
| 6 | `guardedInput` pressed **once**. FE8 drops input during window fade-ins, so a lost press was lost for good and the run wedged in `target_selection` for the rest of the scenario. | Both stuck cases had the battle forecast up and never left `target_selection` |

Waits are now **progress-based** where a fixed budget cannot work: a running event engine is productive work and does not burn the stall budget, with `STALL_CEILING` still failing closed.

## Recovery from a wedged screen

Nicolas's suggestion: if the driver is parked somewhere it never wanted, back out instead of dying. Built with guards, because a *blanket* rescue is exactly what let six defects hide behind the old cadence:

- the cancel is a **legal enumerated action** — `controller.lua` now exposes `cancel_target`, so a selector we cannot commit is escapable at all;
- every recovery is **traced**, so a real defect cannot hide behind it;
- the allowance is small, so a genuinely stuck screen still fails closed.

Honest note: **it has not fired in a passing run.** It is insurance, not a fix — the retry (#6) is what actually turned those scenarios green.

## The bug my own tests missed

`harness.lua` is one Lua chunk and its main function sits near Lua's **200-local ceiling**. Five new top-level constants pushed it over and the whole file stopped loading — every scenario dies at once, and no source-text assertion notices. Nicolas caught it watching a run; my `loadfile` check had been piped to `head` without asserting, so `nil, err` looked like success.

Tuning now lives in one `TUNE` table, and `test_playtest_harness.py` **compiles** `harness.lua`, so `make test` catches this before anything reaches the emulator.

## Verification

`make matrix SUITE=gate`, clean run:

```
canonical  controller_turn    PASS     13s     ch04boot   ch04moose          PASS     13s
canonical  win                PASS     13s     ch04boot   ch04packmath       PASS     19s
canonical  gameover           PASS     13s     ch04boot   ch04village        PASS     10s
canonical  retreat            PASS     13s     ch04boot   ch04cottage        PASS     13s
canonical  ch01               FAIL     31s     ch04boot   ch04snag           PASS     16s
canonical  ch01win            PASS     46s     ch04boot   clear_ch04_parley  PASS     37s
canonical  lordfloor          PASS     22s
testch     recordunitlist     PASS     34s     14 scenarios, 3 builds, 0 duplicate builds, 5m55s
```

Also green: all 28 `tools/test_*.py`, all 8 Lua suites (132 controller assertions), `make check` = `drift check: clean`, `git diff --check` clean.

## Still red — `ch01`, and what it needs

It clears ch00, rides the post-chapter save prompt, then **parks inside the Beat-1 Northlook cutscene** until the flow budget runs out. The event engine stays live but no page wait is ever classified as `dialogue_wait`, so nothing advances it. `ch01win` survives the same scene only because it predates the controller contract and rides it on a blind A cadence.

Fixing it properly means identifying the scene-text input wait and classifying it. I could not pin it down reliably: the freeze report's proc naming is currently untrustworthy (three distinct scripts all resolve to `E_FACE`). **This is precisely what #222's state-inspector workstream is for**, and it is the first concrete argument for taking that workstream rather than deferring it.

Note `reachCh01Map` is shared by `smoke_ch01`, `clear_ch01` and `fuzz_ch01`, so this one blocks more than the single gate row.

### DIFF
--- a/tools/playtest/controller.lua
+++ b/tools/playtest/controller.lua
@@ -248,6 +248,10 @@ function M.legalActions(observation)
         end
     elseif state == "target_selection" then
         add(actions, "confirm_target", "A", "target.currentTarget", observation.target.uid)
+        -- B backs out of target selection to the command menu. Enumerating it is what
+        -- lets a driver RECOVER from a selector it cannot commit instead of wedging
+        -- there for the rest of the run (#232).
+        add(actions, "cancel_target", "B", "target.selection")
     elseif state == "player_map_idle" then
         addCursorActions(actions, observation.cursor)
         if observation.cursor then
--- a/tools/playtest/gen_symbols.py
+++ b/tools/playtest/gen_symbols.py
@@ -65,6 +65,10 @@ WANTED = [
     'GameIntroHealthSafetyWaitButton', # exact health/safety input wait
     'Title_IDLE',               # exact title-screen input callback
     'ProcScr_SaveMenu',         # save/New Game menu
+    # The BETWEEN-CHAPTERS save prompt is a SECOND proc script (savemenu.c) sharing the
+    # same slot-select callback. Watching only ProcScr_SaveMenu left it unclassified, so
+    # the ch00 -> ch01 flow parked on it forever (#232).
+    'gProcScr_SaveMenuPostChapter',
     'SameMenu_CtrlLoop',        # exact save-menu input callback
     'SaveMenu_SaveSlotSelectLoop', # exact New Game save-slot input callback
     'ProcScr_NewGameDifficultySelect', # New Game difficulty menu
--- a/tools/playtest/harness.lua
+++ b/tools/playtest/harness.lua
@@ -31,10 +31,26 @@ local Recorder = dofile(PLAYTEST_DIR .. "/recorder.lua")
 --   src/proc.c sProcArray: 64 procs x 0x6C; +0x00 = proc_script
 local UNIT_SIZE = 0x48
 local US_DEAD = 4 -- include/bmunit.h (1 << 2)
--- How long one attack may take before the combat wait gives up. Measured: an UNSKIPPED
--- ch00 boss animation is 1238 frames, and vanilla's worst cases (crit, a double, a
--- level-up, a promotion) run well past that -- see chooseAttack (#232).
-local COMBAT_WAIT_FRAMES = 3600
+-- Harness tuning, in ONE table rather than one local apiece: harness.lua is a single Lua
+-- chunk and its main function sits near Lua's 200-local ceiling, so a handful of new
+-- top-level locals stops the WHOLE FILE loading and every scenario dies at once (#232).
+local TUNE = {
+    -- How many times one guarded input may be re-pressed into an unchanged, still-legal
+    -- state before failing closed. FE8 drops input during window fade-ins, so a single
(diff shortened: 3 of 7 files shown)


NicoviGH:
**CI note — the `checks` job's red is a GitHub Actions outage, not this branch.**

Four attempts on 2026-08-06 all died in *Set up job*, before a single step ran, with two signatures:

- `Failed to resolve action download info. Error: Service Unavailable`
- `The job was not acquired by Runner of type hosted even after multiple attempts` — cancelled at exactly **15m03s**, which is GitHub's runner-acquisition timeout. Worth flagging: that precision reads like a configured job timeout and invites a wild-goose chase. It isn't one.

The `build` job — which compiles the full ROM and runs `make test` — **passes** on the same commits (8m52s). So CI is not telling us anything about this branch either way.

Local verification on the merge commit: all `tools/test_*.py`, 8 Lua suites (132 controller assertions), `make check` = `drift check: clean`, `git diff --check` clean, and `make matrix SUITE=gate` live at 13/14.

Re-run with `gh run rerun <run-id> --failed`.

Issue #231, opened 2026-08-06
Playtest matrix runner — one command runs the live regression matrix (#222 workstream 1)


NicoviGH:
Child of #222 (workstream 1). Scope agreed with Nicolas 2026-08-05: **workstream 1 only** — workstreams 2–4 (state inspector, declarative scenario manifests, static lint) stay deferred and get re-scoped from the ch05 build experience.

## Why now

ch05 (#25) will run the live matrix repeatedly. Today that means knowing, per scenario, which `make` flag builds the right ROM and which `PT_HOST_CHAPTER` hosts it — facts that live only in `harness.lua` comments. Getting the flag wrong fails the scenario for a bogus reason; getting the order wrong rebuilds ROMs and re-earns checkpoints (`ckpt_ch02start` replays the whole ch00→ch01→ch02 chain).

## Design

One manifest, two consumers.

- **`tools/playtest/matrix.yaml`** — one row per `harness.lua` scenario: ROM config flags, `PT_HOST_CHAPTER`, checkpoint, fps/vsync class, deadline. Plus named suites. Every scenario is enumerated, so "absent from the manifest" is a drift failure rather than a silent default.
- **`tools/playtest/matrix.py`**
  - `resolve <scenario>` — shell-eval'able assignments, consumed by `run.sh`.
  - `run --suite <name> | --scenarios a,b,c` — group by ROM configuration, build each configuration **at most once**, run scenarios sequentially, print a compact `variant | scenario | verdict | duration | artifacts` table, write `results.json`, exit nonzero if any build or scenario fails.
- **`run.sh`** — its `case` blocks (checkpoint map, fps/vsync/deadline policy) are deleted and replaced by a manifest lookup. Behavior unchanged; `run.sh` keeps owning "run one scenario in mGBA".
- **`check.py`** — new lint: every `scenarios.X` in `harness.lua` has a manifest row, and every suite member resolves.
- **`make matrix [SUITE=...]`** — default suite `gate`.

Ordering inside a ROM configuration: no-checkpoint scenarios first (a fast failure surfaces before a long checkpoint build), then checkpoint-backed ones grouped so each `.ss` is earned once. mGBA runs stay sequential (shared emulator/save state); ROM builds stay sequential (two builds in one tree corrupt each other).

## Definition of Done

- [ ] `matrix.yaml` covers every scenario in `harness.lua`
[...]

PR #237, opened 2026-08-06
Name the state instead of guessing at it: playtest inspector + first-divergence report (#236)

### LINKED ISSUE #232: ch01 parks in the Beat-1 cutscene (last of the 7 gate failures)

### LINKED ISSUE #236: Playtest state inspector + first-divergence report (#222 workstream 2)
Child of #222 (workstream 2). **Promoted from "deferred" on evidence**: #232 was six defects that each cost a full build-and-run cycle to find, and the seventh (`ch01`) is still open purely because we cannot see what FE8 is waiting on.

## Why this is now worth buying

The #232 session is the argument. What it actually cost:

- **Every hypothesis cost a ROM build + an emulator run** (~1–8 minutes each) because the only way to ask "what is the game doing right now" was to edit `harness.lua`, rebuild, re-run, and read a log.
- **Two hypotheses were flat wrong** and were only disproved after a full cycle each — "the battle is stalled waiting for input" (it wasn't, it was 38 frames over budget) and "the ch01 cutscene is slow" (it wasn't, and the fixed loop-exit frame I read as evidence meant nothing).
- **`ch01` remains unfixed** specifically because observation ran out: the event engine stays live but no page wait is ever classified, and `freezeReport`'s proc naming is untrustworthy — three distinct proc scripts all resolve to `E_FACE`, so the dump cannot say which proc is actually waiting.

The matrix runner (#231) made *running* the matrix cheap. This makes *diagnosing* it cheap. Those are the two halves of the same velocity problem, and the second half is now the binding constraint.

## What is not obvious going in

Carried here from HANDOFF so it lives with the work, not with live state:

- **The concrete blocker is proc identity.** `freezeReport` names a proc by nearest-preceding
  symbol, so three distinct proc scripts all print as `E_FACE`. That single defect is why #232's
  last failure is still open — the dump cannot say which proc is waiting. Exact-match resolution,
  and an honest "unknown script" when nothing matches, is worth more than any amount of extra
  formatting.
- **`classify()` returning `transition` is the real blind spot.** Five of #232's six defects were
  input waits nothing had a name for; each read as a passive transition and cost a full
  build-and-run cycle to identify. Making *"this is an UNCLASSIFIED wait, and here is what it was
  compared against"* a visible output is the highest-value part of this issue.
- **Reuse the controller observer.** `observeController()` already builds the state; the inspector
  formats and diffs it. A second state model is explicitly out of scope (#222 guardrail).
- **Do not re-run a scenario to re-test a hypothesis the evidence already killed.** A
  budget-bounded loop exits at the same frame no matter what is on screen, so an identical frame
  number proves nothing. That mistake is part of what motivated this issue.

## Scope

Per #222 workstream 2, reusing the controller observer — **do not build a second state model.**

- **One-shot snapshot**, human-readable plus the underlying structured JSON: live classification, chapter/turn/faction, cursor unit + selectability, active semantic menu commands, targets, dialogue wait, Preparations state, and relevant Proc ownership.
- **Trustworthy proc identity.** The current naming is the concrete blocker: resolve a proc script address to the *right* symbol (exact match, not nearest-preceding), and say plainly when an address matches no known script. An unnamed-but-honest proc is far more useful than a confidently wrong `E_FACE`.
- **Name the unclassified.** When `classify()` returns `transition`, report *why* — which live procs were considered and which known state each failed to match. Every #232 defect except one was an input wait nothing had a name for; the inspector should make "this is an unclassified wait" a visible, actionable output rather than something inferred from a hang.
- **Trace diffing**: compare two runs, or a run against an accepted trace, and report the first meaningful classification/action/postcondition divergence.
- **No new runtime.** Raw proc inventory only when a state is unsupported or malformed.

## Definition of Done

- [ ] A developer can inspect live FE8 state without reading raw memory or screenshots
- [ ] Proc identity is exact, and unknown scripts are reported as unknown rather than mislabelled
- [ ] A `transition` classification can explain what it rejected and why
- [ ] A failed regression reports its first semantic divergence with before/after evidence
- [ ] Output is stable enough for CI artifacts and issue comments
- [ ] Pure formatting/diff logic has unit coverage (no emulator in the tests)

## The acceptance case

**Use `ch01` as the proving run.** It parks inside the Beat-1 Northlook cutscene with the event engine live and no page wait ever classified. If the inspector cannot tell us what that scene is waiting on, it has not earned its keep — and if it can, it closes the last #232 failure as a side effect.

Note `reachCh01Map` is shared by `smoke_ch01`, `clear_ch01` and `fuzz_ch01`, so this blocks four scenarios, not one.

Reference: the semantic readouts in [GBA Fire Emblem for Screen Readers](https://github.com/StanHash/GBA-Fire-Embem-for-Screen-Readers), adapted for debugging rather than copied as a second runtime.

Part of #222. Blocks the remaining half of #232.

NicoviGH:
Diagnosing a playtest failure cost a ROM build and an emulator run per hypothesis, because the only way to ask "what is the game doing right now" was to edit `harness.lua` and re-run. #231 made *running* the matrix cheap; this makes *diagnosing* it cheap.

**Gate: 13/14 → 14/14 — the first fully green gate.**

## What it does

| | |
|---|---|
| **Exact proc identity** | `gen_symbols.py` emits `procscr.lua` (952 script addresses → exact symbols) + `symbols.json` (every ROM code symbol, for naming idle callbacks). Exact match only; an unmatched address prints `unknown@0x…`. |
| **Named unclassified waits** | `classify()` is an ordered rule table; `explain()` runs the same rules to return the verdict **plus** every rule considered and the predicate that failed. |
| **Snapshot + stall detector** | `INSPECT.snapshot` dumps that reasoning with the proc inventory, sampled twice so FROZEN is distinguishable from moving. `INSPECT.watch` arms only on an *unclassified* transition holding a byte-identical proc pool. |
| **Render + first-divergence diff** | `inspect_state.py render` / `diff`. Pure logic, unit-tested, no emulator. |

## The `E_FACE` defect was worse than reported

`#232` said three scripts resolve to `E_FACE`. In live capture the `PROC_NAME` field is not just ambiguous, it is **stale**:

| what was running | what the old report printed |
|---|---|
| `gProcScr_Talk` | `E_FACE` |
| `ProcScr_StdEventEngine` | `MAPTASK` |
| `gProcScr_TalkSkipListener` | `ekrBattleEnding` |

Identity now comes from the script address. This is what unblocked everything below.

## The acceptance case, and two things it disproved

`ch01`'s real blocker took one run to name: **`gProcScr_YesNoChoice` in `YesNoChoice_Loop_KeyHandler`** — lord select's *"Will \<lead\> lead the party?"*. It runs *inside* a live event scene, so the `std_event` passive rule swallowed it and nothing ever answered. Classified as `yes_no_choice`, ordered above the passive rules; the controller says what is legal, the scenario picks the answer.

Both of these contradict what the issues recorded, and both are in `decisions.md`:

- **#232's stated cause for ch01 was wrong.** Not "no page wait is ever classified as `dialogue_wait`" — page waits were classified and advanced **74 times, all passing**, and the flow was still advancing **263 frames** before the old 12000-step cap expired. A cap that fires mid-scene reports a timeout that names nothing. Caps are now sized above any real scene (`TUNE.bootSteps`); `INSPECT.watch` is the failure oracle.
- **The boot flow read 74 pages of text it never looks at**, at normal typewriter speed. `reachCh01Map` now sets `gameSpeed` only, leaving combat animation alone.

## Two things worth a reviewer's eye

- **`tools/playtest/test_*.py` was gated by nothing.** Not `make test`, not `check.py` (globs `tools/test_*.py`), not `unittest discover -s tools` (the directory is not an importable package). The pure-logic tests this issue asks for ran only by hand. Both globs now include it, which picks up the pre-existing `test_make_gif.py` too.
- **`harness.lua` is at Lua's 200-local ceiling.** I crossed it mid-change and the whole file stopped loading. The inspector therefore ships as one `INSPECT` table rather than three top-level locals, netting two slots back.

## Verification

- `make matrix` → **14/14**, 3 builds, 5m34s
- `make test` (now including the playtest Python tests) → green
- `make check` → `drift check: clean`
- `git diff --check` → clean

Closes #236. Closes #232.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ campaigns/*/map_sprites/.base/
 /references
 tools/emulator/
 tools/playtest/symbols.lua
+# the other two gen_symbols.py outputs: exact proc-script names for the Lua side,
+# every ROM code symbol for inspect.py (#236). Regenerated after every make.
+tools/playtest/procscr.lua
+tools/playtest/symbols.json
 # local session scratch tool (HANDOFF references it; intentionally unversioned)
 tools/key_magenta.py
 # mGBA save-state checkpoints (per-ROM-build, regenerated) -- inline comments don't work
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -140,6 +140,7 @@ file lean (operating instructions + pointers, not a fact store).
 | `tools/build_campaign.py` | Inject campaign content (portraits, names, character class/stats) into the decomp before `make`. |
 | `tools/verify_text.py` | Decode message text from the built ROM (regression gate; no mGBA). |
 | `tools/playtest/run.sh <scenario>` | Automated in-emulator playtest of ONE scenario (mGBA Lua scripting; memory-asserted PASS/FAIL). Takes the scenario's ROM flag, host chapter, checkpoint and timing from `tools/playtest/matrix.yaml`. |
+| `tools/playtest/inspect_state.py` | Read a playtest log semantically: `render` prints what FE8 was doing and why the controller classified it that way (procs named by exact symbol); `diff` reports the first divergence between two runs. |
 | `make matrix [SUITE=…]` | The live regression matrix (`tools/playtest/matrix.py`): builds each ROM configuration at most once, runs its scenarios, prints a verdict table. `SUITE=gate` (default), a chapter suite, or `all`. |
 | `tools/portrait_tool.py`, `tools/ref_to_bust.py` | Bust art pipeline (ref → indexed FE8 portrait). |
 | `tools/setup-toolchain.sh` | One-time macOS toolchain setup (brew deps, agbcc, python numpy/pillow/pyyaml). |
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ verify:
 # Run the Python unit tests (combat math + stat resolution + difficulty engine).
 # Also run by `make check` / CI / the pre-commit hook.
 test:
-	@for t in tools/test_*.py; do echo "== $$t =="; python3 $$t || exit 1; done
+	@for t in tools/test_*.py tools/playtest/test_*.py; do echo "== $$t =="; python3 $$t || exit 1; done
 	@if command -v lua >/dev/null 2>&1; then \
 		for t in tools/playtest/test_*.lua; do echo "== $$t =="; lua $$t || exit 1; done; \
 	else echo "== skipping Lua playtest tests (no 'lua'; brew install lua) =="; fi
--- a/docs/adding-a-chapter.md
+++ b/docs/adding-a-chapter.md
@@ -141,6 +141,31 @@ Once the chapter has more than a couple of scenarios, drive them together:
 make matrix SUITE=chNN     # one CHNNBOOT=1 build, every chNN scenario, one verdict table
 ```
 
+## When a scenario fails, read the state before you rebuild
+
+Every terminal controller failure and every proven stall dumps an inspector snapshot into the log
+(#236). Read it first — it usually names the defect outright, and it costs no build:
+
+```sh
+tools/playtest/inspect_state.py render /tmp/playtest-<scenario>/playtest.log
+tools/playtest/inspect_state.py diff /tmp/playtest-<good>/playtest.log /tmp/playtest-<bad>/playtest.log
+```
+
+`render` prints the verdict, the rule that produced it, **every rule that was rejected and why**, and
+the live procs named by exact script symbol with their idle callbacks resolved. A verdict flagged
+`*** UNCLASSIFIED WAIT ***` means FE8 is waiting on something the controller has no name for: add
+the proc + its input callback to `gen_symbols.py`, `CALLBACK_NAMES`, `observeController` and a
+`classify` rule, and let the *scenario* decide the answer. That is the whole fix for a wait, and it
+is how ch01's lord-select Yes/No prompt was closed (#232).
+
+Two traps this replaces, both of which cost real sessions:
+
+- **Do not re-run a scenario to re-test a hypothesis the evidence already killed.** A budget-bounded
+  loop exits at the same frame no matter what is on screen, so an identical frame number proves
+  nothing. Instrument for the answer instead.
+- **A `transition` is not "nothing is happening".** It is the classifier saying it has no name for
+  this. The snapshot's `considered` list tells you what it looked for.
+
 ## Gotchas (learned the hard way)
 
 - **Tutorial-list terminator is per-chapter typed.** Slot 4's `EventListScr_ChM_Tutorial` is an
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -3273,6 +3273,38 @@ session state. `HANDOFF.md` points here._
   `run.sh` refuses a mismatch in 0s with the exact `make` line. So this should no longer be a
   judgement call — but `MX_SKIP_ROM_CHECK=1` disables the guard and puts you straight back here
   (which is exactly what happened once inside the session that built it).
+- **A proc's identity is its script ADDRESS, never its `PROC_NAME` string** (2026-08-06, #236).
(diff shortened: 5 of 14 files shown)


NicoviGH:
**CI status: no run scheduled — GitHub Actions hosted-runner outage, not this branch.**

Every run on `main` for the last hour+ fails identically, including commits that touch only `HANDOFF.md`:

```
The job was not acquired by Runner of type hosted even after multiple attempts
checks: .github#1
build:  .github#1
```

(e.g. run 31122529160 on `main` @ `52d054a`.) No workflow run was created for this PR at all. Same condition HANDOFF flagged for #235 — **do not re-diagnose.**

What ran locally instead. Note the CI `checks` job is deliberately a lightweight drift guard (no ROM build — it needs the gitignored base ROM and the agbcc toolchain), so local verification here is a strict superset of it:

| gate | result |
|---|---|
| `make matrix` (SUITE=gate) | **14/14 PASS**, 3 builds, 5m34s — first fully green gate |
| `make test` | green, now including `tools/playtest/test_*.py` |
| `make check` | `drift check: clean` |
| `git diff --check` | clean |

Worth re-running CI before merge once runners recover.

PR #251, opened 2026-08-08
ch05: give Basil and Sahnar identities, and Sahnar the anim we picked (#25)

NicoviGH:
Both shipped their art in #179/#181 and then sat inert. Art without a `PORTRAIT_MAP` slot is art nothing can address: no name, no bust, no stat line, no map sprite, no death quote — and, the thing actually blocking #25, **no pid for a Talk to name**. This lands the identity half. The ch05 event wiring that consumes it (Basil's green Talk, the Basil→Sahnar parley) is the next PR.

### What's in it

| | |
|---|---|
| **Slots** | Basil → `Artur`, Sahnar → `Marisa` — the Trex/Lupin rule (a vanilla character absent from ch00-08, referenced nowhere else). Sahnar's is an exact class match and vanilla's *other* red→blue Myrmidon parley. |
| **Donors** | Basil → Natasha, Sahnar → Joshua. This is where the healer split actually lives: the two Priest YAMLs show identical class data because they **are** the same class — only `STAT_DONOR` makes Basil the frail mage-healer against Sclorbo's durable Moulder. |
| **Faction** | Sahnar gets `recruit.initial_faction: red`, so `recruit_initial_faction` returns RED and the existing faction-parameterized recruit path covers her with **no new machinery**. |
| **Pre-recruit palettes** | Both. A charId-keyed cast override is unconditional, so without these Sahnar wears her finished slate *while hostile* — and FE reads that greyed look as "already acted" (the Trex bug, via Lupin's fix). |
| **Sahnar's battle anim** | The Specter we chose, vendored verbatim from FE-Repo and imported **whole** — 12 modes, 101 frames, native palette. Nothing faked. |
| **Basil's battle anim** | Live on `clone_from: bishop` — the row whose own comment already named him as its reuser. No new donor work. |

### Verification

Verified from the **injected data**, not a screenshot:

- `gPreRecruitVariant` carries `{CHARACTER_ARTUR, 105, …}` and `{CHARACTER_MARISA, 106, …}`
- banim rows land at `0xDB` (basil) / `0xDE` (sahnar); SMS rows at 100/101 + 105/106
- stat lines resolve off their donors (`basil -> Artur: CLASS_PRIEST base<-NATASHA`, `sahnar -> Marisa: CLASS_MYRMIDON base<-JOSHUA`)

ROM builds green · `verify_text` 3404 messages, 0 runaway · `make matrix` **15/15** · `make test` 802 OK · `check.py` clean.

### Three ADRs for what this cost to learn

1. **A dead message id is proven by USE, not by an empty body.** Two death-quote ids were owed and `PC_DEATH_QUOTE_MSGS` carried a TODO to auto-allocate "from a free pool" — there is no pool, and both obvious audits call Lupin's *shipped* `0x974` "used" (`msg.h` `#define`s every id; bare hex collides with addresses). Also argues against the TODO: a floating id would move `verify_text` baselines.
2. **FEditor's `L` is an authoring bracket, not an instruction** — no loop opcode exists in `banim_code.inc`; vanilla encodes the same shape flat. First surfaced here because the Specter is the first vendored anim to use it.
3. **A donor row is completed for the CLASS, not for the unit that lands it** — Sahnar's import doesn't use `motion`/`cadence`, but the row is class-keyed, so `sword` was authored off FE8's own `banim_myrm_sw1` rather than left `None`.

### Needs your red-pen

The two `death_quote:` lines are **drafts** — every other cast quote was co-written with you. Both are marked `DRAFT` in the YAML with the reasoning.

Part of #25 (checklist item "Basil / Sahnar STAT_DONORs + live `battle_anim:`").


NicoviGH:
## All three sprites, both units, captured in-engine

Everything below is a screenshot or a GIF off a real `TESTCH=1` ROM.

review sheet

### Battle anims

| Sahnar — the Specter, native palette | Basil — heal-cast on the Bishop donor |
|---|---|
| sahnar | basil |

### No cadence mixing — proven, not asserted

Diffed the generated `banim_sahnar_motion.s` against Alexsplode's own `Sword.txt`:

- commands emitted that are **not** in his script: **NONE**
- commands in his script that we **dropped**: **NONE**
- my `sword` cadence macros appearing anywhere in it: **NONE**

She fights on the Specter's own timing and its own sounds (C20/C22/C23/C24/C28/C38/C65). The `sword` cadence I added serves the *faked 3-pose generator*, a path her import returns before reaching — it exists only so a future faked Myrmidon doesn't inherit a hole, and it's read off FE8's own `banim_myrm_sw1`.

### Stat lines match the donor math

Sahnar reads Str 8 / Skl 13 / Spd 14 / HP 24 on the status screen — exactly class base + Joshua's personal (9 SKL + 4, 9 SPD + 5, 16 HP + 8).

### One caveat on the review sheet

Portraits and joined map sprites are ROM truth. The **pre-recruit swatches are not** — they're the build-derived sheets under a faction ramp I approximated in the review script. Only ch05, once the recruit wiring lands, will show the engine's real green/red.

### One thing worth your eye

Basil's pre-recruit GREEN puts his body on the faction ramp while his leaves sit on the neutral grey-green — so as an NPC he reads greener and flatter than his joined look. Deliberate (the body is what should change side), but if you'd rather the leaves stay vivid I can move which index carries the ramp.

PR #261, opened 2026-08-11
Stage Ravisin's ch05 eruption warning

### LINKED ISSUE #260: Ch5: allocate host-block IDs and stage Ravisin’s eruption warning
Child of #25.

Ravisin’s approved portrait is now live (#258/#259), so stage the locked turn-2 eruption warning on the real ch05 host.

## Scope

- Allocate named ch05 message IDs from its host-owned free range `0x9E4..0x9F3`.
- Treat YAML labels `vanilla 0x9C5` and `vanilla 0x9C8` as anatomy/source references only. Literal `0x9C5` is already ch04’s Status objective and is not a writable destination.
- Register the new IDs in `HOSTED_CHAPTER_MESSAGE_IDS`.
- Add/extend collision and ownership tests before production wiring.
- Emit Ravisin’s four locked eruption boxes from `ch05-the-elven-tomb.yaml`.
- Add the `TEXTSHOW` beat to the turn-2 wave script without changing the approved staging.
- Run focused tests and `python3 tools/verify_text.py`.
- Record/prove the smallest affected scene on the real ch05 map; do not run the full matrix gate.

## Locked text

1. “Sephek was right. ...They've come further than I expected.”
2. “I better make sure they don't get to the reliquaries...”
3. “No matter. The dead will destroy them first.”
4. “...And the blade under the stone... I'll use her too.”

## Out of scope

Ravisin’s death quote is the immediate follow-up after this beat is visible, but it should use another allocated ch05 host-block ID rather than expanding this slice before the warning is proved.

Closes only this focused child issue; parent #25 remains open.

NicoviGH:
## Summary

- allocate host-owned message `0x9E4` for Ravisin's locked four-box turn-2 eruption warning
- register the ID in `HOSTED_CHAPTER_MESSAGE_IDS` and inject the warning before Sahnar rises
- add focused ownership, box-shape, ordering, harness, and matrix coverage
- correct stale ch05 documentation around vanilla anatomy IDs, Frostmaiden wording, and the retired crest
- fix `recordCutscene` so function terminals observe a dialogue wait before the recorder advances it

## Why

The chapter YAML's `vanilla 0x9C5` label describes the source scene's anatomy. Literal `0x9C5` is already ch04's Status objective, while hosted ch05 owns the Ch6 block. Without an explicit host allocation, wiring this line would silently overwrite another chapter.

The real-map proof initially false-failed because the recorder consumed `dialogue_wait` before invoking the terminal callback. Moving that observation first lets the proof remember that the warning appeared while preserving guarded dialogue input.

## Player impact

Turn 2 now visibly tells the player that Ravisin's dead are racing for the reliquaries, then stages Sahnar's rise from the final locked box. Later reinforcement waves remain silent.

## Verification

- `PYTHONPYCACHEPREFIX=/tmp/manchego-stars-pycache make check` — pass; drift clean
- `python3 tools/verify_text.py` — 3404 messages, 0 runaway
- `python3 tools/verify_text.py 0x9E4` — exact four-box message
- `python3 tools/playtest/matrix.py run --scenarios recordch05eruption` — PASS; four guarded dialogue advances
- `git diff --check` — pass

Closes #260

### DIFF
--- a/campaigns/rime-of-the-frostmaiden/chapters/ch05-the-elven-tomb.yaml
+++ b/campaigns/rime-of-the-frostmaiden/chapters/ch05-the-elven-tomb.yaml
@@ -501,7 +501,8 @@ enemy_units:
 #   0x9C4 ( 4)     on-map    other recruit's beat before Talk    the MOOSE triggers the fight
 #                                                                 (position reused, not content)
 #   ══ everything above is ONE pre-map event list (98 boxes); below is triggered ══
-#   0x9C5 ( 3)     on-map    mid-battle ESCALATION               the eruption
+#   0x9C5 ( 3)     on-map    mid-battle ESCALATION               the eruption (4 boxes;
+#                                                                 hosted at 0x9E4)
 #   0x9CC (32)     on-map    TALK-RECRUIT scene                  Basil Talks Sahnar (payoff)
 #   0x9C6 ( 1)     quote     ESCORT-NPC death quote (Natasha)    Basil's death quote
 #   0x9C7 ( 1)     quote     BOSS TAUNT — one box                Ravisin taunt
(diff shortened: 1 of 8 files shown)


NicoviGH (commented):
Independent pre-merge review of `b3ae574..c8d60ec`: no Critical, Important, or Minor findings. The host-owned `0x9E4` allocation, collision registration, four-box Ravisin emitter, turn-2 ordering, silent later waves, focused tests, recorder regression fix, and stale-documentation corrections all align with #260. Fresh local gates and both GitHub checks are green. Ready to merge.

PR #280, opened 2026-08-14
Wire ch05's scene 5 -- Basil trundles up and joins (#25)

NicoviGH:
Scene 5 of #25's table, worked top to bottom in player order. Host ids `0x9EE` (locked) and `0x9EF` (the no-Lupin arm). **11 done, 6 left.**

## Scene 5 does not inherit its twin's POSITION

`decisions.md` → *"A cutscene's CHANNEL is inherited from the twin"* got the channel right and implied the placement: vanilla plays `0x9C2` **before** its prep `CALL`, so ours would too. It can't.

Vanilla `LOAD1`s Eirika's group onto the street first. **Our party is placed BY prep** — on a prep chapter the ally table is never `LOAD`ed. So at the point the twin plays, ch05's field holds sixteen risen dead, Ravisin, and one green shrub, and Basil's *"Oh! Tourists. In the tomb."* is addressed to an empty pocket **in both arms**.

The beat therefore plays after `CALL(prep)`, in vanilla's own after-prep shape — `FADU(16)` (the shared prep prologue fades to black and leaves it there) → `CUMO_CHAR` → `STAL` → `CURE` → `TEXTSTART` — which is exactly what its own `0x9C3`/`0x9C4` are.

Two things fall out, both good:
- **The ask and the flip are one beat.** The `CUSA` was already after prep for an unrelated reason (Basil is green across the prep screen so Pick Units never sees her). Box 3 is *"...Take me to her?"*; the next command answers it. Before this the join was silent.
- **`branch_on_check_alive`'s `label_base` is used in anger** — two branches now share one event list, so the arrival keeps 0/1 and the join takes 2/3. `BEQ`/`GOTO` scan the list for a matching `LABEL`.

## A fallback line chosen as PROSE has not been BOXED

Three of the five no-Lupin substitutes don't fit one box at the bubble's 29. Scene 5's is 74 characters and, left flowed, paged itself **mid-clause** — *"You just-- came"* / *"here. On your own."* — an A-press the author never placed, on a scene whose locked arm was hand-boxed to this exact width in July.

`variant_beat` now accepts a `script:` entry that is a **LIST** of boxes. `boxes:`/`replaces:`/`script:` still agree one-for-one, so this stays one mechanism. Substitutions resolve against the original beat and splice afterwards — editing in place would shift every later index and the anchor assertion would then blame the locked script for moving.

Nicolas chose the break (2026-08-14): **after the shock**, so her run-on then arrives whole — her register per `lore/basil.md` §Voice, corpus twin Ewan. The no-Lupin arm is 4 A-presses against the locked arm's 3; arms need not match, only stand up.

**The Talk recruit's fallback overruns identically and takes the same treatment when it is wired.**

## Proven in-engine, on both arms

The box count is the discriminator, so a film cannot pass on the wrong arm:

| ROM | scenario | verdict | boxes | arm |
|---|---|---|---|---|
| `ch05boot` | `recordch05join` | PASS | 4 | no-Lupin |
| `ch05lupinboot` | `recordch05joinlupin` | PASS | 3 | Lupin's |

Both terminate on **Basil turning BLUE**, not on "dialogue happened" — the film ends only once the beat has reached its own point.

no-Lupin arm

Lupin's arm

A separate film from `recordch05opening` rather than an extension of it: the two sit on opposite sides of Preparations, and re-filming four proven backdrop scenes to reach a fourth-quarter beat would spend the whole record budget on footage #279 already shipped.

## Also

- Extracts `ch05_beginning_script` so the opening's **order** is testable without a build — it is the first thing a reader loses, and it is load-bearing three times over.
- Parameterises the scene renderers by channel width; `_ch05_scene_and_variant` now serves scene 4 and scene 5 alike, so the remaining three fallbacks cost a call.

## Gates

`make` green · `verify_text` 3404 messages, **0 runaway** · `make test` green · `check.py` clean.

Closes nothing — #25 stays open with 6 scenes left.


NicoviGH:
## Scene 5 on both arms — for review 📱

**NO-LUPIN arm** — 4 boxes. Basil reads the *party*.

no-Lupin arm

**LUPIN'S arm** — 3 boxes. Basil reads the *wolf*.

Lupin's arm

---

### The one thing I want your eye on

In the no-Lupin GIF, watch the box you chose the break for. **FE8's talk bubble scrolls — it doesn't clear.** So after the A-press on *"...You're none of hers."*, that line stays on screen while *"Not one of you. You just-- came here. On your own."* types in beneath it.

The A-press is real and the pacing is the one you picked. But my ASCII preview showed two clean separate boxes, which isn't how it renders. If it reads as one long run-on instead of shock-then-tumble, switching is a two-line YAML change:

- **A** — break on the full stop instead (*"...You're none of hers. Not one of you."* / *"You just-- came here. On your own."*)
- **B** — trim to a single box (*"...You're none of hers. You came here on your own."*), dropping the `just--` stumble

Everything else is green: `make`, `verify_text` (0 runaway), `make test`, `check.py`, and CI. Not merging until you've looked.

PR #281, opened 2026-08-14
Ravisin raises Sahnar in scene 3; wire ch05's scene 6 (#25)

NicoviGH:
Closes the scene-3 design change and scene 6 on #25's table. **12 done, 5 left.**

## The summon costs no box

Nicolas, 2026-08-14: *"you don't need to even add lines... just add sahnars portrait to the scene."* A new **`enters:`** script directive — the mirror of #279's `exits:` — brings a face up mid-scene with no A-press. Scene 3's **seven locked boxes are untouched**:

```
[OpenMidRight][LoadFace][FID_Riev]
[OpenMidRight]A queen's blade. I have walked past that[LF]
stone a thousand times.[A]
[OpenFarRight][LoadFace][FID_Marisa]        <- the summon, no box
[OpenMidRight]It never once occurred to me she might be[LF]
of use.[A]
[OpenMidLeft][LoadFace][FID_Artur]
[OpenMidLeft]...What will you do with her?[A]
[OpenMidRight]It's not your concern.[A][LF]
Come. There is work before dawn.[A][LF]
Prepare my berries.[A]
[OpenMidLeft]...Yes.[A][X]
```

The craft risk on the record was that a loud resurrection would bury the Basil beat the scene exists for. Staging it instead protects the quiet and sharpens it three ways: box 2 stops being a remark about a stone and becomes an appraisal delivered to her face; Basil's question is asked with her friend standing there; and *"It's not your concern."* is said over her. Sahnar takes **far right** — outside Ravisin, not across from her — because Ravisin holds mid-right through scenes 2 and 3 and must not be evicted mid-scene.

## Scene 6 is a plain on-map bubble, at `0x9F0`

It was the opening's problem child for exactly as long as Sahnar was absent from the field. With her standing at the arena it is vanilla's `0x9C3` down to the camera move — `CUMO_AT(12, 6)` → `STAL/CURE` → the `LOAD1` → `CUMO_CHAR` → text, so the player watches her arrive rather than finding her there. No backdrop, no second `FADU` (scene 5 already paid it).

**It cost one A-press, and no words.** *"No one ever came. I stopped counting somewhere in the middle."* is 60 characters — three lines at the bubble's 29, and a box holds two — so the wrapper was splitting it at `"...somewhere in the" / "middle."`. The break is now authored at the full stop, where the understatement she is budgeted one of per scene gets the press to itself. Same lesson as the reliquary lines and scene 5's fallback.

## Sahnar is Joshua and Basil is Natasha, all the way down

Half of how Joshua plays is a **refusal**. `AI_A_07` is `gAiScript_ActionInRange_ExceptNatasha`, and that carve-out is the only reason a fragile Cleric can walk up to a Killing Edge on the arena tile at all. It does not ride the `.ai` bytes: `AiScriptCmd_05_DoStandardAction` routes through `AiIsUnitEnemyAndNotInScrList`, which tests each candidate's `pCharacterData->number` against a list of character ids — and vanilla's list holds `CHARACTER_NATASHA` literally. Copying the bytes alone left Basil a legal target.

`repoint_escort_safe_ai_list` points that list at our escort at build time. **Safe because it has exactly one client in all of FE8**: `.ai = {0x7,` appears once, on `UnitDef_088B5914`, vanilla Ch5's Joshua — read from decomp HEAD, never the built tree. Guarded both ways: `assert_escort_safe_ai_has_one_client` refuses a second client, the patch hard-exits unless it finds vanilla's form, and `src/cp_data.c` joins `PATCHED_DECOMP_FILES` so it restores each build.

In the built ROM:

```c
u8 CONST_DATA gUnknown_085A8A00[] = { CHARACTER_ARTUR, 0, 0, 0 };

CONST_DATA struct UnitDefinition MS_Ch05Sahnar[] = {
    { .charIndex = CHARACTER_MARISA, .classIndex = CLASS_MYRMIDON,
      .allegiance = FACTION_ID_RED, .level = 6, .xPosition = 12, .yPosition = 6,
      .items = { ITEM_SWORD_KILLER }, .ai = {0x7, 0x3, 0x9, 0x0}, },
```

## The rest of #25's checklist

- **`0x9E4`'s fourth box** was a promise she has already kept. Tense fixed and nothing else — *"the blade under the stone"* is the 9BC/9BD callback the box exists for, and Ravisin still filing a queen under inventory.
- **AI** moves from `aggressive` (pursue/charge) to Joshua's `duelist_hold`. A turn-1 crit-Myrmidon that *hunted* would be a different chapter.
- **The brazier** is re-pointed: Ravisin's instrument, not the eruption's flavour.
- **The re-price turns out to be already paid.** `difficulty.py` `_split_force` buckets a `convertible: true` enemy *before* it reads `arrives_turn`, and turn-1 threat is `line + convertibles` — so the model has priced her as present on turn 1 all along. `make difficulty CH=ch05` is byte-identical: line 15 · reinf 6 · convertible 2 · turn-1 9.9, still **PARITY**. #25 wondered whether the 9.9-vs-9.0 comparison flattered us; it doesn't — vanilla's 9.0 counts Joshua the same way, off its CHAR targets. What was wrong was the YAML's *claim* that Sahnar arrived with the eruption.
- **`DEAD_CONCEPTS`** registers the retired phrasings, so a later pass cannot build scene 6 the backdrop the old note asked for.

## Verification

- `make CAMPAIGN=rime-of-the-frostmaiden fireemblem8.gba` green
- `python3 tools/verify_text.py` — 3404 messages, **0 runaway**
- 425 unit tests, `make check` drift clean
- Injected `MS_Ch05Sahnar` and `gUnknown_085A8A00` inspected in the built tree (above)

**Not yet filmed.** The in-engine review artifact would be `recordch05opening` (scene 3's portrait) and `recordch05join` (scene 5 → 6 after prep) — two runs. Say the word and I'll capture and post GIFs rather than burn them unasked.


NicoviGH:
## Filmed — and the first cut was wrong twice, both caught by looking

### Scene 3: Ravisin's Sahnar, on screen, no lines

scene 3 summon

Trimmed to scene 3 alone (9s) rather than the whole 49-box opening, which #279 already shipped.

### Scene 5 → scene 6: the join, then Sahnar on the arena

join and scene 6

The two beats on the far side of Preparations, 50s. Scenario logs:

```
ch05join: Basil is BLUE after 4 boxes -- the no-Lupin (4-box) arm; the join CUSA landed
ch05join: scene 6 played its 7 boxes with Sahnar RED on the arena -- the summon put
          her there and the bubble had her to anchor to
ch05recruit: Sahnar is RED on (12,6) at turn 1 -- the scene-3 summon put her on the board
```

## Two defects the film found that no assertion could

**1. The raised face was buried.** First cut seated Ravisin on MidRight and Sahnar on FarRight. Podiums are a ladder and neighbouring rungs overlap, with FE8 drawing the active speaker on top — so Sahnar spent the whole scene as a hood behind Ravisin's shoulder. Every id, box count, wrap and podium assertion passed while it did.

For a scene of *speakers* that overlap is harmless and vanilla does it constantly — ch05's own scene 4 seats four across adjacent rungs and each is drawn over the others when its turn comes. A **silent** face never gets that turn. Ravisin moves to `Right`, leaving `MidRight` empty between her and Sahnar on `FarRight` — vanilla's own stable two-face right side (MSG_904, MSG_092C, MSG_0937, MSG_0954).

> My first guard here banned adjacency outright and immediately rejected scene 4, which is shipped and accepted. The rule is about **silence**, not adjacency; there is now a test pinning that speakers on adjacent rungs stay legal.

**2. Two stacked speech bubbles** (Nicolas, on the film). The shipped #279 GIF at the same frame has one clean bubble, so this was ours.

My first diagnosis was wrong: `[SendToBack]` sets `TALK_FLAG_4` → `SetTalkFaceLayer`, i.e. portrait **z-order**, not a window close, and `[OpenX]` is only `SetActiveTalkFace`. Nothing "leaves a window open". The real seam is **`TalkPrepNextChar`** (`scene.c:626`) — whenever the *active* face slot differs from the *speaking* one it calls `ClearTalkBubble()` + `StartTalkOpen()` for the active face. Loading Sahnar mid-message opened a bubble for her, then another for Ravisin.

So a silent **mid-scene** arrival is not something the engine does, and vanilla agrees: every mid-message `[LoadFace]` in the corpus has that face speak *next* (MSG_904, MSG_092C, MSG_095A), and its silent loads are always preloads before the first box (MSG_0954, MSG_095D, MSG_095E).

`enters:` is therefore now **`present:`** — *"this character is on screen for this scene and never speaks"* — rendering through the `preload` path `_script_to_message` already had. The honest name: position isn't expressible, so a directive implying it was lying.

```
[OpenFarRight][LoadFace][FID_Marisa]
[OpenRight][LoadFace][FID_Riev]
[OpenRight]A queen's blade. I have walked past that[LF]
stone a thousand times.[A][LF]
It never once occurred to me she might be[LF]
of use.[A]
[OpenMidLeft][LoadFace][FID_Artur]
...
```

The beat survives because it never rested on the arrival — she is absent through scene 2 and stands at Ravisin's shoulder through all of scene 3, so the **contrast** is the summon. *"It never once occurred to me she might be of use"* is an appraisal to her face; *"It's not your concern."* is said over her while Basil holds her tongue.

## Scenarios that would have failed on success

`ch05recruit` waited on `turn() >= 2 and redSahnar()`. With Sahnar red from turn 1 that is satisfied the instant the counter ticks — **before Ravisin says a word** — so the gate would have reported *"eruption warning showed 0 boxes"* and blamed the warning. Split into two assertions: she is RED on (12,6) at turn 1 (which nothing witnessed before — an empty arena passes every other ch05 scenario), and separately the eruption's four boxes on turn 2.

`recordch05join` terminated on the join CUSA, one command before scene 6, so the new footage was off the end of the film. Extended through her seven boxes.

`recordch05recruit` ended turn 1 and sat through the eruption to reach a Sahnar who didn't exist yet; she's on the arena at turn 1 now, so the Talk is immediately reachable.

## Verification

`recordch05opening` PASS 3m47s · `recordch05join` PASS 4m24s · `ch05recruit` PASS 28s (all on `ch05boot`) · 427 unit tests · `make check` clean · `verify_text` 3404 messages, 0 runaway.

NicoviGH:
### Correction: the scene-3 GIF above was a PRE-FIX capture

Nicolas caught it — the clip posted in the previous comment still showed the two stacked bubbles and no Sahnar. Not a stale cache and not an unfixed defect: a stale **file**.

The trimmed clip got built three times. Two of those runs read `/tmp/playtest-recordch05opening` *before* the staging was re-filmed, and one of them finished last and won the filename. I then `ls`-ed, saw a file, and committed it. I had verified the re-filmed **source frames** by eye and skipped verifying the **artifact** — and checking the inputs is not checking the output.

Rebuilt from the fixed frames and verified by decoding the committed `.gif` itself (198 frames):

scene 3 summon

One bubble, tail on Ravisin, Sahnar at her shoulder from the first frame through *"It's not your concern."* — with Basil asking her question in front of her.

The `ch05-join-and-sahnar-alone.gif` in the previous comment is unaffected: it was built once, from `/tmp/playtest-recordch05join`, which only ever ran once, and the scene-3 staging changes touch message `0x9EB` alone.

PR #282, opened 2026-08-15
Wire ch05's scene 7; dress ch05's moose in the Wyrdeer art (#25)

NicoviGH:
Scene 7 of #25's dialogue-wiring table — **the last beat before turn 1** — plus the moose-sprite bug Nicolas caught while this was being filmed.

## The film

scene 7

Scene 6's tail, then scene 7: Pinky asks, the moose breaks out of its pen, Meesmickle answers, the map opens on the party.

## Scene 7 — two boxes, one id

The beat is a setup and a punchline across a **wordless action**, which normally buys a second hosted message id, and #25's allocation has exactly one spare. It doesn't need one. `stage_break:` renders vanilla's own `[BreakTalk]` (`textdefs.txt` 0x80 0x04 → `scene.c` `case 0x04: LockTalk`): the script acts at the pause and `TEXTCONT` resumes the *same* message. Vanilla splits `MSG_9BF` this way; our lore crawl already shipped the idiom.

What the pause does **not** buy is a scene change — the bubble stays up and the faces stay loaded — so the framing goes before the message, not inside it.

## The charge is net-zero

The pen at `(10,0)` is parity-locked (threat 14.1, cornered against the map's top edge). So the moose lunges to `(10,4)` on screen and a **negative-speed `MOVE`** snaps it back off camera — `Event2F_MoveUnit` short-circuits `speed < 0` to a bare `MoveUnit_`, which is vanilla's own off-screen reposition (`ch14a` resets Carlyle with `MOVE(0xffff, …)` twice around one scene).

Nicolas asked for the charge to *land* on the start tile. The pen is on row 0 — the map's top edge — so there is no tile behind it to charge from; forward-and-checked is the same net zero the other way round, and Ravisin's hold is a check the chapter already describes.

## Two things only a run could say

**`CAMERA` is the scroll; `CUMO` is only a pointer.** `EventDisplayCursor_Loop` calls `PutMapCursor` and returns. The first cut of this wiring used `CUMO_AT` where the party framing was meant, and the film simply never cut south. **Scene 6 had the same gap** and looked right only because `LOMA` happens to leave the camera on the map's north-west, which happens to hold the arena — anything that moved the reload origin would have played it off-screen with no error anywhere. Both beats now name their tile.

**No party unit is named for a camera.** ch05 deploys 9 of a 10-unit pool, so *both* of this scene's speakers can be benched — and a benched unit is still in the array (`US_NOT_DEPLOYED`) at stale coordinates, so `CUMO_CHAR` would pan to nowhere rather than fail. The bubble needs no unit either way: `StartTalkOpen` anchors it to the speaking **face slot**.

## ch05's moose was a dog

Spotted by Nicolas off a playtest frame. Sprite tables are keyed on **pid**, and a pid is per-chapter: `SCRIPTED_NEUTRAL_SPRITES` named ch04's `0xce` alone, so ch05's moose (`0xb9`) matched nothing in `GetUnitSMSId`'s override scan and rendered `CLASS_GWYLLGI`'s stock hound on the red enemy palette — with the Wyrdeer sheets committed and ch05's own YAML claiming they'd shipped a fortnight earlier.

`assert_declared_map_sprites_injected` couldn't see it: it asks whether an **asset** reached a sprite table, and this one had — under ch04's pid.

- A registry row now takes a **tuple of pids**. One sheet pair and one SMS slot per asset; one override row per pid.
- `assert_custom_art_pid_wired` is the per-**pid** half, called by both chapter injectors.

Post-injection (HEAD has neither): `gMapSpriteOverride` `0xb9,107` + `0xce,107`; `gMuImgOverride` both → one sheet; both in `gMapPaletteOverride`.

## Verification

| what | result |
|---|---|
| `tools/test_build_campaign.py` | 462 pass |
| `tools/check.py` | drift clean |
| `make … fireemblem8.gba` | green |
| `verify_text.py` | 3404 messages, **0 runaway** |
| `MSG_9F1` in the ROM | `…running?![03]` **`[80][04]`** `[09]You had to ask?[03]` |
| `recordch05join` (ch05boot) | **PASS**, 4m33s |

One run, extended rather than added — it samples the moose every frame, because *"ended where it started"* is indistinguishable from *"never moved"* unless something watches the middle:

```
[f014048] Basil is BLUE after 4 boxes -- the no-Lupin (4-box) arm; the join CUSA landed
[f015565] scene 6 played its 7 boxes with Sahnar RED on the arena
[f015895] the moose BROKE from its pen -- (10,0) -> (10,4)
[f016075] scene 7 played its 2 boxes; the moose charged and is back on (10,0)
```

## Still owed

The moose's **battle animation** (`battle_anim_todo` on the ch05 YAML: PMD SpriteCollab 0899). Unchanged by this PR — ch04 never needed one, ch05 does, and it is its own piece of work.

Part of #25.


NicoviGH:
## The framing call, in three frames

One open question for you, and it's easier to see than to describe. Base URL is this branch, so these render inline on web and mobile.

### 1. The hold on the pen — Pinky's line

the hold on the pen

The moose is the pale shape at the very top, **half cut off by the edge of the screen**. That's the thing I'd like your call on.

It isn't a camera mistake I can tune away: its pen is `(10,0)`, row 0, and the map sprite is 32px drawn centred on a 16px tile — so its top 8px sit *above* the map. The camera clamps at the map edge, so there is no position that shows row 0 any lower. The only ways out would be moving the pen (parity-locked) or framing somewhere else entirely and losing the "look at the thing he's asking about" shot.

My read: it's fine, and arguably right — *penned against the top edge* is literally the fiction, and the clipping makes it read as trapped. But you own staging.

### 2. The charge landed — and the elk

the charge

`(10,4)`, clear of the top, mid-screen, among the risen. This is also the clearest look at the sprite fix — **a pale Wyrdeer elk with brown antlers**, where before this PR the same tile held `CLASS_GWYLLGI`'s stock hound on the red enemy palette.

Both faces are still up across the `[BreakTalk]`, which is what makes the setup/punchline read as one exchange rather than two scenes.

### 3. Turn 1 opens on the party

turn 1

The `CAMERA(5, 18)` fix. Before it, `CUMO_AT` alone drew a cursor and never scrolled, so the map came up still looking at the north rim.

---

Stills decoded from the **committed GIF**, not from the frame directory — an artifact is not its inputs, and the file is what gets reviewed.

**No before/after of the hound**, deliberately: the pre-fix frames were overwritten by the re-film, and regenerating them costs a build plus a five-minute run to prove something the post-injection tables already say (`gMapSpriteOverride` had `0xce,107` and no `0xb9` row at all). Say the word if you want it anyway.

NicoviGH:
## Both tweaks in, and the beat now has its own boot

scene 7

The GIF is now **scene 7 alone** — no opening, no prep, no join.

### 1. The run is longer, and it turns

Top-right corner → left along the rim → down at the party, as an **authored REDA route** rather than one `MOVE`, because the path is the beat and a pathfinder would cut the corner. `reda_route_move` is lifted out of ch04's flee beat rather than written twice — same animal, same thing, opposite direction.

Row 0's walkable stretch is `x=10..14` (x=1..8 are pillars, x=9 impassable), which is what makes a corner start legal at all.

Still net-zero: two instant `MOVE(0xffff, …)` brackets, one before the camera arrives and one after it has cut south. The fight starts on the parity-locked pen and neither reposition is ever on screen.

the charge

### 2. The bubble comes down

You were right and I'd documented the cause without following it through: `[BreakTalk]` only *locks* the talk proc, so Pinky's box hung over the whole action. `[CloseSpeechSlow]` is `ClearTalkBubble()` and nothing else — faces stay loaded, the talk state survives, `TEXTCONT` brings the window straight back. Vanilla uses it mid-message for the same reason (`MSG_9BF`).

One guard fell out of it: a break between two turns by the **same** speaker is now refused at build time. The bubble reopens on `!TalkHasCorrectBubble()`, which compares the speaking face slot — so same-speaker-both-sides would print into a cleared window and look like the text simply vanished.

### 3. `--ch05-moose` — the boot that should have come first

| | |
|---|---|
| `recordch05join` | **4m33s** |
| `recordch05moose` | **34s** |

New Game → title → chapter intro → the beat. Nothing you've already approved.

Two things learned by getting them wrong: `bootToMap()` is the wrong driver on such a ROM — it drives to `player_map_idle`, and here the beginning scene *is* the beat, so it mashed A through both boxes and handed the film an idle map (3000 frames of nothing). It stops at `chapter_intro_input` now and the film opens on the `FADU`, so the camera move and the hold are in shot. And the flag is one branch at the top of `ch05_beginning_script`, with a test that the shipping opening still carries prep and scenes 5–7.

Written up as `decisions.md` → *"Playtest runs are the most expensive thing in this repo"* rule 3, generalised: `TESTCH` and `--lord-boot` are the same move, and any feature whose screen is hard to reach should get one **before** its first film.

### Verification

`recordch05moose` **PASS**, 34s:

```
[f001644] it RAN the route -- corner, left along the rim, then down to (10,4)
[f001824] 2 boxes; back on (10,0) -- net-zero, so the fight starts where parity says
```

469 unit tests · drift clean · `verify_text` 0 runaway · `MSG_9F1` carries `[15][80][04]` (CloseSpeechSlow, BreakTalk) between the boxes.

### Open

The pen-hold clipping from my last comment is now **moot** — the moose starts in the corner, not the pen, so the opening shot is a different frame. Worth an eye on the GIF anyway.

Moose **portrait + battle anim** still owed and yours in parallel. One note so you don't redo work: `portraits/white-moose.png` (96×80, indexed) already exists and is committed — it is injected *nowhere*, in either chapter, so what's missing there is wiring rather than art. The battle anim is the genuine gap.

PR #283, opened 2026-08-15
The white moose BELLOWS over the map, full screen (#25)

NicoviGH:
Your art, your idea. Scene 7 now reads:

> Pinky: *"...It's not running. Why isn't it running?!"* → **the moose fills the screen, bellowing** → it breaks and runs → Meesmickle: *"You had to ask?"*

scene 7

## Why it's a background, not a portrait

That was your question, and the answer is that it shouldn't be a portrait. A 96×80 bust is drawn inside the talk window's envelope and those antlers were never going to fit it. A `BACG` owns all 240×160 and has **no envelope at all** — so showing the creature properly stops being a compromise. ch05 already flashes full-screen images over the live map four times (the reliquary visits), so the channel was sitting there unused for this.

## How it looks like an *overlay*

FE8 has no image-over-map primitive — a 240×160 image can only live on the BG layers, which is exactly where the map is drawn, and OAM can't hold a full-screen overlay. So:

**the map is in the picture.** The camera is pinned at a known tile for this beat, so the map view at that exact framing was captured and the moose composited onto it. The image then **cuts in with no fade**. The player sees a moose loom over the map with the terrain and the risen still visible around it.

The one real cost: tile animation is frozen for the 90 frames it's up.

## Two things the run said and reading did not

**A locked talk does not survive a scene change.** `[BreakTalk]` only *pauses* the talk proc; `REMOVEPORTRAITS` — which the BACG loader needs re-armed — tears it down, and `TEXTCONT` then resumes nothing. Filmed: the image played, the charge played, and the punchline never appeared. So the beat splits into two messages. `stage_cut:` now sits beside `stage_break:` with the distinction written down — a **break** for anything that leaves the talk state alone (movement, music, a camera hold), a **cut** for anything that changes what's on screen.

**`CLEAN` is the restore, and it isn't optional.** `EventScr_RemoveBGIfNeeded` only fades, conditionally. Without `CLEAN` the map came back wearing the CG's *palette* — a snowfield in moose-red. Vanilla's own order is fade down → clean → fade up.

## Cost

The quip takes `0x9D2`, ch05's one spare id. **The host block is now exactly spent** — the taunt, Basil's death quote and the four ending ids fit with nothing left over. If another beat needs one, re-run the neighbourhood sweep rather than assuming slack.

## Verification

| what | result |
|---|---|
| `recordch05moose` | **PASS**, 37s |
| unit tests | 473 pass |
| `check.py` | drift clean |
| `make` | green |
| `verify_text` | 3404 messages, 0 runaway |

```
[f001818] it RAN the route -- corner, left along the rim, then down to (10,4)
[f002004] 2 boxes; back on (10,0) -- net-zero, so the fight starts where parity says
```

Four iterations, four films, **37s each**, all on `--ch05-moose`. The same work through `recordch05join` would have been eighteen minutes of replayed opening.

## Still owed

The moose's **battle animation** — your three frames are in hand. Note the injector wants `Ready → Windup → Peak`, so your left-to-right (attack, windup, idle) maps reversed; I'll handle that, but I need a `clone_from:` donor class from you.

Part of #25.


NicoviGH:
## Swapped to the tomb plate — you called it

scene 7

the bellow

The baked-map trick was structurally doomed, not just imperfect. It composited perfectly — but the risen standing in that captured frame hold still for the 90 frames the image is up, and the player has *just watched those units move*. A frozen map reads as a photograph of a map.

A scenic plate promises no motion, so there's nothing to give away.

Recorded next to the BG registration as a general rule, because it will come up again:

> **A still image may stand in for a still thing; it may never stand in for something the player has just watched move.**

Framing is unchanged and is the one you signed off — the illustration's flat right edge flush to the frame, 94% height, bottom anchored, so it reads as a portrait bleeding off frame. Still cuts in with no fade.

`recordch05moose` **PASS** (37s) · 473 unit tests · drift clean.

Issue #285, opened 2026-08-16
Parity aggregate should read personal lines on both sides (blocked on the undentable-boss / clear-load problem)


NicoviGH:
The parity AGGREGATE (`threat/slot`, `clear-load/slot` — the numbers the CI gate reads) projects every unit off **class base** and drops personal lines on both sides. The ROLE check was fixed to read the real article in #25 (commit `c8865c6`); the aggregate was deliberately left alone, and this issue is why, plus what it would take.

## Why it should change

FE8 puts a named unit's teeth in its personal line. Measured with personal lines applied to **both** sides, every chapter moves *toward* 1.00 — i.e. the current metric is systematically mis-stating the comparison, not conservatively:

| chapter | threat, class-base (today) | threat, real article |
|---|---|---|
| ch01 | ×1.15 | **×1.08** |
| ch03 | ×1.12 | **×1.03** |
| ch05 | ×1.08 | **×1.04** |
| ch04 | ×1.15 | ×1.15 |

## Why it is not a one-liner

**1. Clear-load goes infinite, on a real FE8 fact.** Vanilla Ch5's Saar with his personal line sits at **Def 13** against the parity yardstick's **13 attack** — exactly zero damage, so `rounds_to_kill` is `inf` and any ratio against it is undefined. This is not a bug to patch out; an early-game iron-sword unit genuinely cannot dent Saar.

The repo already has the mechanism — ch08 reports *"5 yardstick-proof units excluded from clear-load"* — but it has to be applied **symmetrically** and its policy justified, because excluding a unit from clear-load silently makes a chapter look cheaper to clear.

**2. It re-baselines every curated chapter,** including the three `[locked]` ones (CH0/CH1/CH2). Every baseline needs re-approving.

**3. ch02 falls out of band.** On the real article its clear-load drops to **×0.64**, below the 0.75 floor. That is either ch02 genuinely under-loaded — likely, and see #284, where its boss turns out to be a naked class base and its clear-load already sits exactly on the floor at ×0.75 — or an artifact of how exclusions land. **Resolve #284 first**; it may move ch02 enough that this stops being a blocker.

## Definition of done

- [ ] Personal lines applied to both sides of the aggregate, from both sources (`personal:` in chapter YAML, and `BASE_DONOR` for a cast member deployed hostile — see `unit_real_article()`)
- [ ] Undentable units handled symmetrically, with the exclusion policy stated and tested
[...]

Issue #284, opened 2026-08-16
ch03's boss is a naked class base; ch02's only looked like one (the tool could not see slot-inherited lines)


NicoviGH:
> **Corrected 2026-08-16 after measuring.** This issue opened on the premise that ch02's and ch03's bosses were both naked class bases. **Only ch03's was.** ch02's boss deploys on a vanilla character slot the build never patches, so the ROM had been giving it the real line all along — the "1.2 rounds" was the measuring tool, not the game. The body below is the corrected statement; the original framing is preserved at the bottom. Fixed in #287.

A named boss in FE8 is **class base plus a personal stat line** — vanilla Bazba is a L6 Brigand *plus* HP+5/Pow+3/Skl+4/Spd+2/Def+2/Res+2/Lck+1, and that line is most of why he reads as a wall. Surfaced while balancing ch05's white moose (#25), from a warning `role_findings()` had been printing all along.

## What was actually true

| boss | claimed | measured | why |
|---|---|---|---|
| ch02 `raider-captain` | naked, 1.2 rounds | **3.6 rounds — at the bar, always was** | deploys on the **BAZBA slot**; nothing patches it, so FE8 adds Bazba's own line in the built ROM |
| ch03 `grell` | naked, 1.1 rounds | **confirmed naked** → now 3.6 | rides raw pid `0xb7`, and a CharacterData gap is all zeros |
| ch05 `ravisin` | reference | 13.4 (bar 12.9) | the pattern both were measured against |

## The real defect was in the measurement

A personal line reaches a unit **three** ways. `unit_real_article()` knew two: `personal:` in the chapter YAML (Ravisin), and `BASE_DONOR` for a cast member deployed hostile (Sahnar on Joshua's). The third — **an enemy deployed on a vanilla CHARACTER slot** — had no representation at all, so every such unit was projected off naked class base.

Verified against the **built** character table, not by reading the injector:

| our unit | slot | after injection | measured |
|---|---|---|---|
| ch01 `goblin-chief` | BREGUET | unchanged | 5.2 → **6.2** (bar 6.2) |
| ch02 `raider-captain` | BAZBA | unchanged | 1.2 → **3.6** (bar 3.6) |
| ch02 `raider-bruiser` | BONE | unchanged | — |
| ch00 `sephek-kaltro` | ONEILL | **zeroed by the build** | 2.1 (bar 2.2) |

[...]

PR #294, opened 2026-08-19
Three durable findings get a home in decisions.md (#25)

NicoviGH:
Written into `HANDOFF.md` during this session and all true in three sessions, so they belong here and get pointed at from there. (The handoff-side trim lands on `main` after this merges — `check_handoff_only_on_main` correctly refused to let me stage it on a branch.)

- **The TESTCH bench's real ceiling.** It ran out at 7 and I wrote down "the bench is full", which invites a redesign. The 7 is one row's geometry (y=9, spacing 2 so the bait cannot counter a neighbour, 15-wide map); a second row lifts it. The actual ceiling is SMS VRAM — `ResetUnitSprites` hands out `0x40` = 64 slots, `gSMS16xGfxIndexCounter` walking **down** 1 per distinct 16×16 and `gSMS32xGfxIndexCounter` walking **up** +2/16×32 and +4/32×32 until they **cross** and later sprites silently overwrite earlier ones (what `recordunitlist` fails on). Cost is per **distinct sprite**, not per unit, so today's bench spends ~10 of 64.
- **`recordenemy`'s shared output path**, cleared per run — folded into "An artifact is not its inputs", whose "one output path, one owner" half this is. It does not take two jobs racing; two in sequence will do it, and it cost a re-run of a scene Nicolas had already watched. The same ADR now also records that a **repair must be verified against the ORIGINAL**, never against the damage the previous repair left — which is exactly how Ravisin's map sprite ended up with holes through her face.
- **A lock has a date** (#293): scene 17 called Basil "he" because the text was locked 2026-07-30 and her `gender: female` was settled 2026-08-08. "Locked" was read as "checked", which it never was — and the exposure grows with the gap, which matters because ch05's endings were locked in July and ship in August.

Docs only; drift check clean.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4596,6 +4596,55 @@ ramp entries into a single colour destroys form either way. Ship the intent with
 (`docs/demo/ch05-ravisin-palette.png`). Judge it offline first — this whole loop cost seconds and no
 ROM build, the same reason `rom_bg_preview.py` exists.
 
+### A LOCK has a DATE, and facts settled after it still apply (2026-08-19, #25, #293)
+
+ch05's scene 17 called Basil **"he"** twice. Her text was locked 2026-07-30; her `gender: female`
+was settled 2026-08-08. Nobody contradicted anybody — **the lock simply predated the decision**,
+and "locked" was read as "checked", which it never was.
+
+So a locked scene is frozen against *re-litigation*, not against *facts*. Before wiring one, diff
+its lock date against anything settled since about the characters it names — pronouns, class,
+who is recruitable, who is even alive. Correcting a scene to match a later-settled fact is not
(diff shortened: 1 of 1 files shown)


NicoviGH:
Closing as ceremony I should not have added. `decisions.md` lands *inside* the feature PR that produced the decision here (#289, #285, #284, #286 all did), and the only standalone `docs:` commits — 3f12324, 5ae30d2, 0d90716 — go straight to `main`. This is docs-only housekeeping cleaning up my own misfiling, so it went to main as 40ff12c, together with the `HANDOFF.md` trim that had to live there anyway.

Splitting it across a PR made it worse rather than safer: it forced a two-step where the handoff half could not travel with the ADRs it points at.

Issue #303, opened 2026-08-22
Support all three difficulty modes (RULED: vanilla ships three, so do we)


NicoviGH:
Nicolas asked, 2026-08-21: *"Do we support the three difficulty modes?"* The measured answer is **we inherit them, we do not support them, and nothing verifies them** — and the first part of this issue is a decision only he can make.

## What is true today, verified

- **The difficulty select screen is LIVE.** New Game passes through `ProcScr_NewGameDifficultySelect`; the playtest harness just presses A on whatever is default. A player can pick Difficult.
- **FE8 implements difficulty as a per-chapter enemy LEVEL SHIFT** (`easyModeLevelMalus` / `normalModeLevelMalus` / `difficultModeLevelBonus` in `chapter_settings.json`) plus `extraTrapsInHard`. There are no per-difficulty enemy TABLES — enemies come from our own event `LOAD`s, so our units are fielded in every mode. That part is fine.
- **We point `playerUnitsInNormal` AND `playerUnitsInHard` at the same ally table.** Correct: our cast does not change.
- **We never set the level shifts.** Each chapter inherits its donor host slot's numbers, tuned for a different chapter:

| ours | host slot | easy | normal | difficult |
|---|---|---|---|---|
| ch01 | 1 | −0 | −0 | **+1** |
| ch03 | 5 | −2 | −0 | **+3** |
| ch04 | 5 | −2 | −0 | **+3** |
| ch05 | 6 | −4 | −2 | **+3** |

  ch03 and ch04 share host slot 5, so they **necessarily** share difficulty numbers. That cannot be right for both.
- **`tools/difficulty.py` does not model modes at all**, and every playtest scenario runs one configuration. So every PARITY verdict this project has produced is *Normal-only* — which is not what the reports say.
- **Some content is mode-gated and we already rely on it.** `CHECK_TUTORIAL` is `!(gPlaySt.config.controller) && !(chapterStateBits & PLAY_FLAG_HARD)` (`eventscr.c:834`), and ch05's arena tutorial is wrapped in `EventScr_CallOnTutorialMode` exactly as vanilla's is. So on Difficult, that tutorial does not play — deliberately, but nobody has looked at it.

## The decision, first

**Do we ship three difficulty modes at all?** Three honest options:

1. **Support them.** Level shifts become declared per chapter, `difficulty.py` grades the mode, and at least one gate runs on Difficult.
[...]

Issue #241, opened 2026-08-06
Code-review follow-up on the #237/#239/#240 stack


NicoviGH:
Two reviewer passes over the merged stack (`71b37de..c489c56`). One finding breaks CI the moment the Actions runners recover; the rest are landmines for ch05's scenarios.

## Critical

- [ ] `check_hosted_chapters_declared` (`tools/check.py:199`) imports `build_campaign`, which imports Pillow at module scope. The `checks` job installs **only pyyaml** and runs `python3 tools/check.py`, so `main()` exits 1 on every push — with a misleading message. Simulated: `["build_campaign does not import: No module named 'PIL'"]`. It also contradicts the rule at `tools/check.py:660` ("read the constant names out of the tuple rather than importing build_campaign"). Fix: dependency-free hosts module that `build_campaign` re-exports.

## Important

- [ ] Local-slot headroom is documented as 2 in all three places (`harness.lua:276`, `check.py:169`, `HANDOFF.md:74`); measured it is **1** (+1 compiles, +2 fails). #240 consumed a slot and updated no comment. Make the number executable so it cannot go stale.
- [ ] `dialogue_wait` sits above `yes_no_choice` (`controller.lua:134` vs `:146`) — if both are ever live, A gets consumed by the choice handler and #232 silently returns. Untested pairing.
- [ ] The stall signature is blind to `PROC_REPEAT` (`harness.lua:1034-1058`): `scrCur`/`idleCb`/`lockCnt` are constant for a whole scene under it, and `p.sleep` is excluded even though `INSPECT.snapshot`'s own `frozen` compares it.
- [ ] `bootToMap` still uses the literal `12000` cap (`harness.lua:1097`) that `TUNE.bootSteps` exists to retire.
- [ ] `LUA_CHUNKS` (`check.py:138`) lists 4 of the 9 chunks the harness loads — glob it.
- [ ] `hosted_chapters()` misses the prologue (`PROLOGUE_HOST_INDEX` does not match the regex), so a future `CH05_HOST_INDEX = 1` collides silently; enrolment is convention, not a gate.
- [ ] `INSPECT.watch` — the new failure oracle — has zero tests because it lives in `harness.lua`. Moving it to `controller.lua` tests it and buys back the local slot.
- [ ] A missing `symbols.json` degrades silently (`inspect_state.py:51`); every callback renders `unknown@...`, indistinguishable from a real unknown.

## Minor

[...]

PR #304, opened 2026-08-22
difficulty: declare all three modes per chapter, don't inherit them (#303)

NicoviGH:
Closes the mechanism half of #303. **We ship three difficulty modes on purpose now, with declared numbers instead of whatever host slot we squatted.**

## What FE8 actually does

It authors **one** enemy table per chapter and derives all three modes from it by re-projecting stats at unit-load time (`UnitApplyBonusLevels`). Three 4-bit fields per chapter are the entire difference between the modes. We never wrote them — so every hosted chapter carried its donor slot's numbers, tuned for a different chapter.

Each chapter YAML now declares its triple, `apply_chapter_difficulty` writes it, and a chapter that declares none **fails the build**.

The write is its own registry pass, not a line inside `_retarget_host_chapter` — the prologue never retargets, so that placement would silently skip a chapter (#241's shape). Mirrors `CHAPTER_BATTLE_TILESETS`.

## Two corrections to what the issue assumed

- **ch03 and ch04 do not share a host slot.** `hosted_chapters()` raises if two chapters claim one; the real slots are 1/2/3/4/5/6. The founding case as written could not happen.
- **FE8's modes are Tutorial / Normal / Difficult**, and `easyModeLevelMalus` is the *Tutorial* slot. Menu option 0 sets `controller=0/HARD=0`, which is both the `-easyModeLevelMalus` branch (`eventscr.c:2328`) and the only state `CHECK_TUTORIAL` fires in — so **ch05's arena tutorial is absent on Normal too**, not just on Difficult. Worth a separate call on whether that is what we want.

## The defect this fixes

**ch04's Normal was out of band.** It hosts on slot 5 (`I05`, normal malus 0) while FE8 Ch4 carries 2:

```
ch04 normal inherited  verdict=OFF  (threat 9.18 vs 7.05)   x1.30
ch04 normal twin       verdict=OK   (threat 8.23 vs 7.05)   x1.17
```

ch03's Difficult drifts +3 to +2. ch00/ch02/ch05 were already correct, by luck of squatting the right slot.

## Why the twin's numbers keep the tuning we already did

Both sides of a parity read are **authored** tables — `difficulty.py` reads level straight from the YAML and straight from the decomp's `UnitDefinition` arrays, unshifted. So our table sits on the same footing as vanilla's, and vanilla's own spread lands our three modes where its three sit. Measured, with the shift modelled on both sides:

| chapter | authored | tutorial | normal | difficult |
|---|---|---|---|---|
| ch00 | x1.13 | x1.13 | x1.13 | x1.07 |
| ch01 | x0.89 | x0.89 | x0.89 | x0.90 |
| ch02 | x0.81 | x0.81 | x0.82 | x0.81 |
| ch03 | x1.03 | x1.02 | x1.03 | x1.03 |
| ch04 | x1.15 | x1.17 | x1.17 | x1.08 |
| ch05 | x1.04 | x1.01 | x1.06 | x1.06 |

Max drift from the authored ratio: **0.04**. The shift is parity-preserving.

`mode_stats()` is faithful to the engine's two non-obvious behaviours: the penalty **floors** at `baseLevel` (`UnitAutolevelPenalty` re-derives from base and skips the re-autolevel), and the Difficult bonus **rounds twice**. It honours the `>= 0x3C` gate too — vanilla Ch5's pre-recruit Joshua (0x20) is difficulty-immune, as is our Sahnar on Joshua's own slot. One unshifted unit per side, mirroring.

**Stated ceiling:** `GetAutoleveledStatIncrease` is random (jitter plus a coin-flip remainder, `bmbattle.c:1241`), so no static model predicts a given unit's stats. `mode_stats` is a mean-value proxy — but run through one function on both sides, which is what makes the comparison valid.

## Playtests were grading Tutorial

The menu initialises to option 0 and the harness only ever pressed A — so **every verdict this project has produced graded the easiest mode** while reading as general. `PT_DIFFICULTY` now names the mode and the controller navigates to it on the yes/no-choice contract (A commits what is highlighted, so the wrong option must be moved off first). `INSPECT.difficultyProbe` plus the `difficulty` scenario read the red force's stats *on the map*; the menu only proves we clicked something.

## Verification

692 Python tests, 260 Lua assertions, `check.py` clean. All TDD — every test watched failing first.

Worth noting: `check.py` caught two defects in this very branch — the new test classes sat after `unittest.main()` and would never have run in CI, and two new top-level locals had consumed harness.lua's last free slots against Lua's 200-local ceiling.

⚠️ **In-engine run still pending** — building the ROM and running the probe in all three modes next. Not merging until that lands.

## Left

- `difficulty.py --mode` on the CLI (model built and tested, flag unwired)
- Whether ch05's arena tutorial should stay Tutorial-only now that we know Normal loses it as well


NicoviGH:
## In-engine verification complete, and it found a real defect

Ran the mode probe on ch05 in all three modes. First pass showed the modes **do** adjust (red maxHP 368 / 390 / 440) — but with the two named bosses **inverted**, weakest on Difficult.

### Root cause

`UnitAutolevelPenalty` fires only `if (level > pCharacterData->baseLevel)`. **Every vanilla named boss ships `baseLevel` >= its deploy level** — Saar 8/8, Breguet 4/4, Bazba 6/6, Novala 10/7, Murray 12/9. That is how vanilla protects a hand-authored boss line so the same stats reach the map in every mode.

Our four bosses on RAW pids sat in `gCharacterData` gaps where `baseLevel` reads 1, so the penalty always fired — resetting them to class base and rebuilding from growths, discarding the `personal:` line their YAML authored. The two promoted ones were loudest: the reset path re-runs full `UnitAutolevel` (promoted branch, +9 levels) while the Difficult path calls `UnitAutolevelCore` direct and never grants it.

The community documents the same routine from the other side — when the penalty target falls at/below `baseLevel` the recalculation is skipped entirely, the known weak-promoted-enemies bug in the localized ROM ([FEU thread](https://feuniverse.us/t/fe7-fe8-difficulty-stat-changes/1295)).

### Fix and result

`RAW_PID_LEVEL_SOURCES` writes `baseLevel` = the deploy level, read from the chapter YAML so it cannot drift. Measured in-engine:

| unit | before (tut/norm/diff) | after |
|---|---|---|
| Ravisin | 39 / 40 / **35** | 34 / 34 / **35** |
| white moose | 28 / 29 / **23** | 21 / 21 / **23** |
| generic | 20 / 22 / 26 | 20 / 22 / 26 (untouched) |
| Sahnar (0x16) | 24 / 24 / 24 | 24 / 24 / 24 (immune, as expected) |

34 is Ravisin's authored line exactly (class base 19 + personal 15). Tutorial and Normal now ship it untouched; Difficult is strongest. Totals 356 / 378 / 440.

Guarded by `unregistered_raw_pid_bosses()` — the failure is silent, since a gap is all zeros and nothing complains.

⚠️ **Every playtest before this graded Tutorial**, so ch03's and ch05's bosses were measured in their reset-and-rebuilt form, not their authored one.

Also landed: `difficulty.py --mode`. 701 Python tests, 260 Lua assertions, `check.py` clean.

Issue #302, opened 2026-08-22
Epic: chapter as code — make a chapter cost fewer EVENINGS, then make it DECLARED


NicoviGH:
> **Epic.** Re-scoped 2026-08-22 after measuring where ch05's effort actually went. The original
> premise (a chapter should be DECLARED, not copied) still holds and is still the endgame — but it
> was aimed at the wrong cost. See "What the measurement changed" below.

## The goal, in Nicolas's words

*"With the number of chapters we've done by now we should be able to have a model for 'chapter as
code' so to speak, where you already know how to build and wire everything, you know all the
elements of a chapter."* (2026-08-21) — and the point of that model is **to make our efforts less
laborious going forward** (2026-08-22).

**"Less laborious" is measured in evenings and watched playtest runs, not in commits or constants.**

## What the measurement changed

The first framing counted `CH0N_*` constants (37 → 48 → **78** across ch03/ch04/ch05) and concluded
the injector was the problem. Then we counted where ch05's **102 first-parent commits** actually
went:

| | share |
|---|---|
| `docs/` + `HANDOFF.md` **only** | **50 commits (49%)** |
| content-touching | 31 (30%) |
| tooling, no chapter YAML | 17 (17%) |
| outright fix / rework | 4 (4%) |

By lines added: injector + its tests + playtest scaffolding = **69%**, against **3.4%** for the
chapter YAML.

Two facts follow, and they point at different work:

1. **The 20:1 ratio is real.** Every unit of declared content drags ~20x its own weight in Python.
   That is what the declarative model fixes.
2. **Half the commits were the session boundary.** The log shows the cadence directly:
   `docs: point the next session at ch05's scene 5` → `Wire ch05's scene 5 (#280)` →
   `docs: point the next session at ch05's scene 6` → … Fifteen scenes, one per session.

And (2) has a cause: **verification is a watched GUI run.** `tools/playtest/run.sh` line 2 —
*"Run ONE automated playtest scenario in mGBA (**headed**, but fully scripted)"*. So:

```
verification is a watched GUI run  →  batch size = 1 scene per session
                                   →  ~15 changeovers
                                   →  50 handoff commits (49% of ch05)
```

Theory of Constraints: every system has one constraint, and optimising a non-constraint adds
nothing to throughput. **The constraint is verification cost measured in Nicolas's attention.**
`inject_ch05` is not the constraint. It is merely the visible thing.

[...]

PR #316, opened 2026-08-23
playtest: verdict scenarios run HEADLESS (#308)

### LINKED ISSUE #308: playtest: run verdict scenarios HEADLESS, so a run stops costing Nicolas's attention
Child of #302. **The constraint**, per the ch05 measurement: verification is a *watched GUI run*, which forces batch size to one scene per session and produced 50 handoff commits (49% of ch05).

`tools/playtest/run.sh` line 2: *"Run ONE automated playtest scenario in mGBA (**headed**, but fully scripted)"*. It launches `mGBA-dev.app/Contents/MacOS/mGBA` — the Qt GUI binary.

## Spiked 2026-08-22 — it works

mGBA 0.11 ships its own headless frontend (`src/platform/headless-main.c`, `BUILD_HEADLESS`, off by default) with the **full** script context: `--script`, `mScriptContextRegisterEngines`, `AttachCore/Stdlib/Image/Logger/Storage`.

Neither mGBA's macOS nightly DMG nor pokeemerald-expansion's vendored `mgba-rom-test-mac` will do: the DMG ships only `mGBA.app`, and the vendored binary advertises `--script` but has **no Lua compiled in** (their tests are C-in-ROM). It has to be built:

```sh
brew install cmake libzip lua
git clone --depth 1 https://github.com/mgba-emu/mgba
cmake -B build -DBUILD_HEADLESS=ON -DBUILD_QT=OFF -DBUILD_SDL=OFF -DUSE_LUA=ON
cmake --build build          # -> build/mgba-headless
```

Configure reports `Headless: ON · Scripting: ON · Lua 5.5.1 · Screenshot/savestate: ON`.

**Proven headless on this Mac:** frame callbacks fire, `emu:addKey`/`clearKey` round-trip (START = 8), the game executes, and `ch03prep` booted to chapter 4, drove the dialogue (`advance_dialogue … "result":"pass"`) and reached *"Preparations screen opened; Fight! into the phase"* — with no window.

## The one blocker, root-caused

`emu:screenshot()` **segfaults** headless: `mCore_screenshot` → `PNGWritePixels` → `KERN_INVALID_ADDRESS at 0x0`. The headless frontend attaches no video renderer. `shot()`'s `pcall` cannot catch it — a C segfault is not a Lua error.

That is not a workaround to find, it is the **designed split**:

- `kind: verdict` → **headless**, no screenshots (they assert on memory: `INSPECT.units`, `activeMsg()`)
- `kind: record` / `diagnostic` → stay **headed**, screenshots intact — this is how art and feel defects get caught (#279's letterbox bar and Sahnar's exit were eyes-on finds)

`matrix.yaml` already types every scenario with exactly that `kind` field, so the switch keys off data that exists.

## Scope

- [x] Vendor/build `mgba-headless` into `tools/emulator/`, same auto-fetch shape as the nightly
- [x] `PLAYTEST_HEADLESS` guard in `harness.lua` `shot()` — log the intent, skip the call
- [x] `run.sh` selects the binary from the resolved `kind`; `PT_HEADED=1` escape hatch
- [x] ~~A guard: a `kind: verdict` scenario that calls `shot()` is a drift failure~~ — **dropped, and it was wrong.** `shot()` degrades to a logged no-op under headless, so a verdict scenario calling it is harmless; `ch03prep` legitimately calls it three times and passes. A guard here would fail correct code.
- [ ] Re-baseline the gate headless and record wall-clock vs the 24m51s headed number

NicoviGH:
Closes #308. First child of the re-scoped #302 epic to land.

## Why this, first

#302 originally measured `CH0N_*` constants and concluded the injector was the problem. Counting where ch05's **102 first-parent commits** actually went says otherwise:

| | share |
|---|---|
| `docs/` + `HANDOFF.md` **only** | **50 (49%)** |
| content-touching | 31 (30%) |
| tooling | 17 (17%) |
| rework | 4 (4%) |

Only four rework commits — ch05 wasn't bug-plagued, it was *slow*. The cause is upstream:

```
verification is a watched GUI run  ->  batch size = 1 scene per session
                                   ->  ~15 changeovers
                                   ->  50 handoff commits
```

Theory of Constraints: optimising a non-constraint adds nothing to throughput. The constraint is verification cost measured in Nicolas's attention.

## What changed

- `kind: verdict` → **`mgba-headless`, no window**. These assert on memory (`INSPECT.units`, `activeMsg`) and never needed pixels.
- `kind: record` / `diagnostic` → **stay headed on purpose.** Their output *is* the picture, and eyes-on is how #279's letterbox bar and Sahnar's exit were caught.
- `matrix.yaml` already typed every scenario with `kind`, so this keys off data that already existed — no new hand-kept field.
- `PT_HEADED=1` forces headed. Falls back to headed with a note if the binary isn't built, so a fresh checkout still works.

## Verified

All PASS, headless, no window, on the ch03boot ROM already in the tree:

| scenario | time |
|---|---|
| `ch03prep` | 14s |
| `ch03chest` | 14s |
| `ch03talk` | 21s |
| `ch03midmap` | 21s |

`make check` clean. The headed comparison was **deliberately not run** — re-running green scenes to obtain a number is exactly the cost this PR exists to remove.

## Three traps, each recorded in `decisions.md`

1. **Nobody ships a macOS binary with both headless and Lua.** mGBA's nightly DMG has only `mGBA.app`; pokeemerald-expansion's vendored `mgba-rom-test-mac` advertises `--script` and has **no Lua compiled in**. Hence `tools/build_mgba_headless.sh` — which verifies Lua actually loads afterwards, because a Lua-less binary builds fine and still accepts `--script`.
2. **`emu:screenshot()` segfaults headless** — `mCore_screenshot` → `PNGWritePixels` → null deref. That is a C segfault, so `shot()`'s existing `pcall` is no protection at all; the call must not happen.
3. **`PT_HEADED` had to join `PLAYTEST_ENV_KEYS`** or a headed PASS could be served for a headless run. `test_playtest_matrix` caught this — the guard working as designed.

## Review notes

- `PLAYTEST_HEADLESS` is read as a **global**, not bound to a top-level local: `harness.lua` is one chunk against Lua's 200-local ceiling with 2 slots free. My first version spent one of them.
- A guard rejecting `shot()` inside verdict scenarios was drafted and **dropped as wrong** — the skip degrades gracefully, and `ch03prep` calls `shot()` three times and passes.
- `AGENTS.md` is edited on disk to match `CLAUDE.md` but deliberately **not committed** (untracked by design, per HANDOFF).


NicoviGH:
## Review pass — six findings, all fixed in 2c6167b

Reviewed at `high`. **The first pass should not have merged**: it carried a latent checkpoint-corruption bug.

### HIGH — checkpoint builders inherited the headless engine

`run_mgba` read a single `$APP` global set from the *scenario's* kind, so a headless scenario's checkpoint builder ran headless too. And `emu:saveStateFile()` is broken there — verified directly:

```
saveStateFile returned: false
file size: 397312
all zeros: true
```

The builder would still PASS on its own assertions, `run.sh` would stamp `.romhash` **valid**, and the dead state would look fresh forever. Exposed set, enumerated from the manifest — `ch02`, `ch02baxby`, `clear_ch02`, `smoke_ch02`, all on `ch02start` — with `tools/playtest/states/` **empty**, so the next run of any of them would have hit it.

Fixed two ways: `run_mgba` takes the engine as an argument and the builder passes `headed` as a literal; and `saveState()` now raises `controllerFault` when the save returns false, so a build that writes nothing reports FAIL instead of PASS.

### HIGH — the build script hung at its last step

The Lua verification ran in the foreground with no timeout, and there is no SWI in a stock FE8 ROM to exit on. Now backgrounded and reaped.

### MEDIUM — `recordsupply` / `recordunitlist` would have lost their frames

Both are `kind: verdict` **by what they assert** (per `decisions.md`), but both drop frames for `make_gif.py`. Inferring headless from `kind` alone silently stripped them.

`headless` is now a **declared manifest field** defaulting to `auto`, and those two pin `headless: false`. Better than the inference it replaces — the manifest says what is true rather than having run.sh guess.

### MEDIUM — silent fallback defeated the cache key

With `PT_HEADED` unset, headed-vs-headless depended on whether the binary existed, which is in no key — the exact collision the key exists to prevent. A missing binary now **refuses in 0s** with build instructions.

### MEDIUM — ~1GB of unread log per long scenario

`mgba-headless` logged every BIOS SWI and DMA: 5.4MB in 6s, captured into `mgba-stdout.log` and never read. `-l 0` → **0 bytes**.

### LOW — `PT_HEADED=0` forced headed

Was a `-z` test; now compared against `"0"`, matching `PT_SOUND` in the same file.

---

**Re-verified:** `make check` clean, 162 matrix tests pass, `ch03prep` still PASS headless, stdout log 0 bytes, states dir confirmed unpoisoned.

**Not verified live:** the checkpoint path is proven by construction plus the new guard, not by a real mint — building `ch02start` is a 900s headed replay of the whole ch00→ch01→ch02 chain, which is precisely the cost this PR exists to reduce.

PR #317, opened 2026-08-23
playtest: verify the checkpoint path for real, and fix what that found (#308)

NicoviGH:
Refs #308. You asked me to actually verify the checkpoint path instead of asserting it was "proven by construction." Doing that found a second bug the first fix did not close.

## What the verification found

`run.sh` classified every verdict with `case "$VERDICT" in *PASS*)` — and `VERDICT` is the whole `RESULT: ...` **line**, reason text included. The guard message I added in 2c6167b read:

> refusing to let a checkpoint build report **PASS** with no state written

So the FAIL it produced **matched the PASS arm.** `run.sh` stamped `prep.romhash` valid off a build that had just written 397,312 bytes of zeros — exactly the poisoning the guard existed to prevent, reached by a different route.

All four globs were fragile, and the last one is the **exit-code line**, so a failure whose reason mentioned the word could have made `run.sh` exit 0.

Reading the code proved the guard fired. It did not prove what the *caller* then did with that verdict. Only running the negative case showed that.

## Fixes

- All four globs match `*"RESULT: PASS"*`; the guard message no longer contains the word.
- A failed checkpoint build removes its partial `.ss`. Without the stamp the next run rebuilds anyway, but a 397KB file in `states/` reads as a real checkpoint to whoever looks next.
- `run_mgba` announces which engine served each invocation. That choice decides whether screenshots and `saveStateFile` work at all, and it was invisible before — it is what made both bugs observable.

## Verified live, both directions, on the canonical ROM

**Negative** — builder forced headless:
```
engine: headless (mgba-headless)
saveState prep -> false
RESULT: FAIL -- ... no state was written
checkpoint build FAILED -- aborting
states/: no .romhash written, partial .ss removed
```

**Positive** — `ch02`, which is headless and depends on `ch02start`:
```
engine: headed (mGBA)            <- the builder, as required
saveState ch02start -> true
RESULT: PASS -- ch02start checkpoint saved

engine: headless (mgba-headless) <- ch02 itself
RESULT: PASS -- ch02 entered: 3 chwinga green, 5 deployed, archer + boss present
```

A real state is 105–129KB at ~99.4% non-zero; the broken one is exactly **397,312 zero bytes**. That size is the tell, and it is recorded in `decisions.md`.

`make check` clean.

## Note on the tree

Verifying this required the **canonical** ROM (all six checkpoint builders run on it), so the tree now holds canonical rather than ch03boot — ROM, ELF and `.build-config.json` all restored from the same cache entry, with symbols regenerated. `states/` now has valid `prep` and `ch02start`, which four scenarios needed and which was empty before.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -2213,6 +2213,31 @@ re-running green scenes to obtain a number is exactly the cost this ADR exists t
    a missing binary must REFUSE, never fall back to headed: which engine ran is not in the verdict
    key, so a silent fallback serves a headed PASS for a headless run.
 
+6. **A verdict glob on the bare word matched a FAIL, and that is how the stamp got written.**
+   `run.sh` classified every verdict with `case "$VERDICT" in *PASS*)`, and `VERDICT` is the whole
+   `RESULT: ...` LINE -- reason text included. The first guard message said "refusing to let a
+   checkpoint build report PASS with no state written", so the FAIL it produced *matched the PASS
+   arm*: run.sh stamped `prep.romhash` valid off a build that had written 397312 bytes of zeros.
+   Four globs were fragile, and the last one is the EXIT CODE line, so a failure could have exited
+   0. All four now match `*"RESULT: PASS"*`, and the guard message no longer contains the word.
+   **This was found only by RUNNING the negative case** -- forcing a checkpoint builder headless
+   and watching what it wrote. Reading the code proved the guard fired; it did not prove what the
+   caller then did with that verdict. A failed build now also deletes its partial `.ss`, because a
+   397KB file in `states/` reads as a real checkpoint to whoever looks next.
+
+   Proof, both directions, on the canonical ROM (2026-08-23):
+
+   ```
+   builder forced headless:  engine: headless -> saveState prep -> false
+                             RESULT: FAIL ... -> checkpoint build FAILED -- aborting
+                             states/: no .romhash written, partial .ss removed
+   normal (ch02, headless):  engine: headed (mGBA)      -> saveState ch02start -> true  -> PASS
+                             engine: headless           -> ch02 PASS
+   ```
+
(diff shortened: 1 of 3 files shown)


NicoviGH:
## Second review pass — both findings fixed, plus the class rather than the instance

Both were in code the first pass added. Fixed in `ec96870`, and each verified against the specific failure it claims to fix rather than by inspection.

### Failed rebuild deleted state it had not written

Correct, and my first test of it hit a *different* case — under headless the builder really does overwrite the state with zeros before failing, so removal was right there. The reviewer's case is a builder that never reaches `saveState()` at all. Now compared by mtime across the build:

```
5s builder deadline (fails without writing)   -> prep.ss byte-identical (58e99b031ebf),
                                                 no cleanup message
builder forced headless (writes 397312 zeros) -> "removing the partial 'prep' state this
                                                 run wrote", no .romhash stamped
```

### Checkpoint abort skipped verdict-cache eviction

Correct. Proved it properly by planting a fake green rather than trusting an empty directory:

```
planted: 1 entry  ->  checkpoint build FAILED -- aborting  ->  after abort: 0 entry
```

### The un-filed suggestion — taken

> hoisting it to a single `VERDICT_PASS=0/1` computed once would close the class rather than the instance

Agreed, and it's the right read: four copies is *why* the previous bug was wrong in four places simultaneously. Now one `verdict_passed()` helper with four call sites, and one place where `RESULT: PASS` appears outside a comment.

### Still green

`make check` clean. Both checkpoints valid — `prep.ss` 105,705 bytes and `ch02start.ss` 128,940 bytes, 99.4–99.5% non-zero, stamped against the tree ROM hash `a44afcc2ac09`.

The end-to-end case from the PR body still holds: `ch02` runs headless while its `ch02start` builder runs headed, both PASS.

Issue #335, opened 2026-08-29
AI is authored, not derived — audit ch00–ch05 against their vanilla twins, then guard it


NicoviGH:
Every chapter picks a **vanilla twin** so its difficulty is grounded rather than guessed. We derive the twin's classes, levels, inventories, item drops and reward vehicles — but **not its AI**, which has been authored by feel. That's an oversight, not a design choice: the AI is in the same `UnitDefinition` structs we already read for everything else.

Found while deriving ch06's roster (#334): the draft declared 13 pursuing units where vanilla Ch6 fields two.

## The audit

Our declared `ai_pattern` (by unit) against each twin's actual `.ai` bytes:

| chapter | ours | vanilla twin |
|---|---|---|
| ch00 | 1 hold, 2 aggressive | 3 distinct, incl. `{0x6,0x3}` DoNothing+NeverMove |
| ch01 | 6 aggressive, 3 hold_then_attack, 1 hold | mixed: 3 pursue, 2 `{0,0,1,0}`, 2 AttackInRange, 1 GuardTile, 1 charge-after-1 |
| ch02 | 5 aggressive, 2 hold | **4 pillage (`AI_B_04`)**, 2 charge-after-1, 2 `0x11`, 1 pursue — vanilla has **no static units**, and we field **no raider** |
| ch03 | 2 aggressive, 8 defensive | 9 AttackInRange + 1 GuardTile — **vanilla has zero pursuers; we ship two** |
| ch04 | **18 aggressive**, 5 defensive | 10 AttackInRange, 10 pursue, 2 charge-after-1, 1 other — vanilla is half static, ours is 78% pursuing |
| ch05 | 10 aggressive, 6 raider, 5 defensive, 1 hold, 1 duelist | 9 AttackInRange, **6 pillage**, 4 pursue, 2 charge-after-1, 1 GuardTile, 1 duelist |

**ch05 is the exception, twice:** its 6 `raider` map exactly onto vanilla's 6 `AI_B_04 PillageThenPursue` (the village-raid race), and `duelist_hold` is byte-identical to vanilla Joshua's `{0x7, 0x3, 0x9, 0x0}`. Where someone reached for the decomp, it is right — which is the proof the rest is reachable too.

The drift runs one direction: **toward aggression**. That converts "advance into fixed threat ranges" — the puzzle FE maps are built on — into a swarm that comes to you.

## Why no gate caught it

[...]

Issue #48, opened 2026-06-19
Difficulty engine → all chapters: enemy-pressure parity (per-chapter + campaign curve)


NicoviGH:
**Decision:** `docs/decisions.md` → "Per-chapter parity beyond Ch1 = enemy-pressure vs a `parity_reference` vanilla chapter". This issue is the design + execution plan. First pipeline-track sub-project of the tooling/delivery roadmap.

## Why
`tools/difficulty.py` currently only grades Ch1 (the lone hand-curated vanilla reference). To make balance a cheap **CI gate for every chapter slice as it's authored** — and to make the 21-chapter back half tractable — the engine must give a trustworthy parity read for ALL chapters. Vanilla *parties* are player-chosen (don't extract cleanly), but vanilla *enemies* do; and since our cast is fixed + at parity (#45), enemy pressure is what sets per-chapter difficulty.

## Design (enemy-pressure parity, base-level first)
- **`parity_reference: "FE8 ChN"`** — new field per authored chapter YAML; single source of truth for which vanilla chapter sets the bar (today only ch01 notes its cadence base).
- **Vanilla enemy extraction** — parse the reference chapter's enemy UnitDefinitions (class/level/count) from the decomp's `chN-eventudefs.h` via `vanilla_decomp_text` (HEAD); stat lines via the existing class-base + autolevel path.
- **Enemy pressure** vs a fixed yardstick unit, per deploy slot: **threat/slot** = Σ(enemy dpr vs yardstick) ÷ deploy cap; **clear-load/slot** = Σ(enemy bulk) ÷ deploy cap. Compare our-chapter vs vanilla-reference; flag outside a band (~±25%, tunable). The yardstick cancels in the ratio.
- **Absolute cast check (kept):** still report our cast (base level) vs our enemies — throughput / durability / carry.
- **Base-level first:** leveled stat projection deferred (fast-follow; needs recruit schedule, #45 item 5).

## Execution checklist
- [ ] Add `parity_reference` to the authored chapter YAMLs (Prologue–Ch8); decide each one's vanilla cadence bar.
- [ ] `vanilla_enemies(parity_ref)` extractor (decomp udef header → enemy stat lines), TDD vs known vanilla data.
- [ ] `enemy_pressure(enemies, deploy_cap, yardstick)` metric (threat/slot + clear-load/slot), TDD with hand oracles.
[...]

PR #339, opened 2026-08-31
ai: borrow every enemy's AI from its vanilla donor (#335)

NicoviGH:
**Stack 1 of 5.** Merge in order; each PR is based on the one before it.

## What

A chapter picks a vanilla twin so its difficulty is grounded rather than guessed, and we already derive that twin's classes, levels, inventories and drops from its `UnitDefinition` structs. The `.ai` field of those same structs was the one we never read — it was authored by feel through six per-chapter `CHnn_AI` label tables.

Because #48 computes threat from stats and weapons, the drift was invisible to every gate: five chapters measured x1.00 and played nothing like their twin.

**Each enemy now names a `donor:` and the build reads that unit's four AI bytes.** No label vocabulary in between — the labels *were* the translation layer the drift lived in (`aggressive` meant one thing in the YAML and another in five injectors; `defensive` and `hold_position` were byte-identical under two names; ch00's `ai_pattern:` was wired to nothing at all). All 63 entries already named their counterpart in a comment, so `donor:` promotes prose that was already there to data the build reads. `ai_override:` with a mandatory `why` is how a chapter declares it means to differ.

## The part worth reviewing closely

**A donor matches where a vanilla unit FIGHTS, not where it is placed.** `xPosition`/`yPosition` is frequently a SPAWN tile; `redas` walks the unit to its post, and the last REDA point is the destination. Matched on spawn tiles, 9 of ch05's 23 units appeared to line up and the nine were coincidences — one paired our bone-archer wave with an ARMOR_KNIGHT boss whose throne AI it would have inherited. Matched on posts it is **23 of 23**, class for class.

## Verification

An **output diff**, not just the tests: every injector re-run and `events_udefs.c` diffed — 102 changed lines, all `.ai`, zero collateral. It caught a bug the whole green suite missed: every injector called `enemy_ai_initialiser(chap, enemy)` without the position index, so all eight of ch05's tomb-reavers shipped their *first* donor's AI. YAML right, resolver right, tests green, ROM wrong. `PerPositionAiReachesTheEmittedRows` pins it.

`make check` green.

Refs #335

### DIFF
--- a/campaigns/rime-of-the-frostmaiden/chapters/ch00-prologue-a-dagger-of-ice.yaml
+++ b/campaigns/rime-of-the-frostmaiden/chapters/ch00-prologue-a-dagger-of-ice.yaml
@@ -177,7 +177,14 @@ enemy_units:
       - "Cold regeneration, fire vulnerability — the Frostmaiden's chill knits his wounds; flame unmakes them (echoed later by the Ch8 ice trolls)"
       - "True nature (book p.23): a drowned mariner possessed by a frost-druid spirit of Auril; kill the body and the winter spirit dies with it — but a single defeat only scatters him"
       - "Misty step: vanishes in a breath of frost — the diegetic reason he escapes the prologue despite being 'defeated'"
-    ai_pattern: hold_position
+    donor: [14, 8]              # O'Neill, the tutorial boss, on this very tile
+    ai_override:
+      ai: '{GuardTileAI, 0x9, 0x20}'
+      why: >
+        O'Neill's own AI is DoNothing ({0x6,0x3,0,0}) and it only works because vanilla's
+        prologue event-scripts his attack -- prologue-eventscript.h CHAIs him to {0,0} once
+        CHECK_TUTORIAL fails. We run no such script, so borrowing his bytes would field a
+        boss that never acts. He holds his tile instead.
     # Defeated on the field, but he does NOT die here — see the chapter_end cutscene (escape).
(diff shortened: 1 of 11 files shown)


NicoviGH:
**Pre-merge review (fresh reviewer, not the author).** Core verified sound: donor
resolution independently re-derived against the decomp for ch01/ch03/ch04, the REDA
destination matching confirmed load-bearing, and `PerPositionAiReachesTheEmittedRows`
reproduces the per-position bug it pins (checked by reintroducing it, not by reading it).
No silent fallback: an enemy declaring neither donor nor override raises.

Two fixes pushed (7798a06):

1. **`ai_override`'s `why` was documented as mandatory and enforced nowhere.**
   `enemy_ai_bytes` returned the vector the moment the key existed; the only check lived in
   `ai_donor_findings`, which the curve gate reads for `balance_locked` chapters ONLY. So
   every chapter still being authored -- where new AI actually gets written -- could declare
   a divergence with no reason and build green. Now raises where the bytes are produced. All
   three existing overrides already carry a `why`, so no byte moved.
2. **`nth:` is gone from the prose.** The previous commit retired the ordinal but left its
   paragraph in `resolve_donor`'s docstring and its name in the "not a coordinate" error,
   including a half-sentence the removal had truncated.

**Known and deliberate:** `make difficulty-gate` exits 1 at this commit, because the new
gate arm hard-fails `balance_locked` ch02 and ch02 is migrated in #340. Verified exit 1
here, exit 0 at #340 and at the stack tip. **#339 and #340 are atomic with respect to CI --
merge them back to back.**

Issue #26, opened 2026-05-31
Ch6 — map + events (The Maer Monster)


NicoviGH:
**The merfolk are the fight. Messie is a cutscene.** He sank Bremen's boats to keep their crews off water the merfolk control — clumsy protection the town read as a monster. Objective is a plain `defeat_boss` against Nerra, the merfolk elder; Messie hauls himself onto the ice once she falls, and Marty and Braulo talk to him. One outcome, no branch flag, no hidden Talk: every player sees the beat instead of the few who would have guessed the command existed on a boss. Canon supports it rather than bending it (Rime pp. 28–31).

Rationale: `docs/decisions.md` → *"Messie is a cutscene, not a boss"* (2026-08-29).

**Design board:** https://claude.ai/code/artifact/6952f53d-0fde-4a0f-b07c-b8fc846d6f10
**Live status:** `make chapter CH=ch06` — derived, never stale. This list is the *decisions*; that command is the *state*.

## Decisions — all settled

- [x] **D1 — donor + bar.** Layout is **Ch13 Ephraim** (22×22, 137 `TERRAIN_RIVER` in concentric channels — the only candidate whose water is a *route*, not a border). The pressure bar is **FE8 Ch6**. Deliberately different chapters; the off-by-one that claimed ch05's bar is fixed. (#331)
- [x] **D2 — boss fight, or the place you arrive at?** Plain `defeat_boss` vs Nerra. Vanilla Ch6 puts its difficulty in a rescue clock rather than its Def-5 boss, and ours inherits that shape: the clock is the two marooned boats. (#334)
- [x] **D3 — the three modes.** `4 / 2 / 3` — vanilla Ch6's own offsets, verbatim. **Fog is NONE in every mode**; the earlier "fog on the harder modes" proposal was rejected, because 40% of this map is water with eight crossings, so the route IS the puzzle and it has to be legible.
- [x] **D4 — deploy cap + start positions.** Cap 10 (Ch6 parity), on Ch13 Ephraim's own eleven start tiles minus its last.
- [x] **D5 — enemy count and composition.** Vanilla Ch6's own twenty-four re-dressed as merfolk, plus its Difficult-only crab-rider trio on a turn-4 load. Every class, level, inventory, drop **and AI byte** is derived from a named donor unit (#335). Parity: **x1.00 threat / x1.00 clear-load**, role check clean. (#334)
[...]

Issue #314, opened 2026-08-23
playtest: scenarios as DATA — a Given/When/Then case, and a chapter declares its own


NicoviGH:
Child of #302. **Borrowed from pokeemerald-expansion's battle test DSL** — and this is the borrow that closes the loop between the two halves of the epic.

Their cases read as declarations:

```c
SINGLE_BATTLE_TEST("Rain multiplies Fire-type moves by 0.5x", s16 damage)
{
    PARAMETRIZE { ... }
    GIVEN { ... }
    WHEN  { TURN { MOVE(player, MOVE_EMBER); } }
    SCENE { ANIMATION(...); HP_BAR(opponent, captureDamage: &results[i].damage); }
    FINALLY { EXPECT_MUL_EQ(results[0].damage, Q_4_12(0.5), results[1].damage); }
}
```

Ours is imperative Lua in a single 9,176-line chunk, already **at the 200-local ceiling** (`decisions.md` → "`harness.lua` is ONE Lua chunk, at the 200-local ceiling").

## Why this is the keystone

If a scenario is DATA, then:

1. Scenario authoring stops being surgery on a file that is out of local slots
2. **A chapter can declare its own scenarios** — which means `matrix.yaml` rows get *generated* instead of hand-written, one of the epic's named derivations
3. `check.py check_playtest_matrix` stops policing hand-sync between two files and starts checking a derivation

## Scope

- [ ] Extract the shape of the existing scenarios — most are boot → drive to a state → assert memory → verdict
- [ ] A declarative case form covering that shape, with an escape hatch to raw Lua for genuine oddities (the DECORATE lesson: code becomes the exception you NAME, not the default you copy)
- [ ] Chapter YAML declares its scenarios; `matrix.yaml` rows are derived
- [ ] Keep `kind` (verdict/record/checkpoint/diagnostic) — it is load-bearing for the headless split (#308) and for `check_verdict_scenarios_are_guarded`

⚠️ Do not port every scenario at once. The wide-edit rule applies: prove it on a handful, diff the OUTPUT, then move.

PR #340, opened 2026-08-31
ch02: restore vanilla's decoy, and stop giving the protectees the best escape (#26)

NicoviGH:
**Stack 2 of 5.** Based on #339 — merge that first.

## What

ch02 is our twin of vanilla's *The Protected*, and it had quietly lost the thing that map is built on.

Vanilla Ch2 fields **six pillaging raiders** and four villages: the villages are the decoy that pulls the raid off the NPCs you are protecting. Our ch02 shipped `villages: []` and three weaponless chwinga standing in the open — so every raider's pillage AI fell through to "attack the nearest thing", and the nearest thing was the people the chapter asks you to protect. It also gave the protectees Mov-7 flight, i.e. the best escape on the field, which is the opposite of vanilla's Ross-and-Garcia problem.

This restores vanilla's shape:

- **The two Targos huts become real villages** — `Village()` events, visit rewards, and the destruction map-change behind them. ch02 shipped two village-terrain cells backing no events at all, so nothing on the map could be visited *or* raided.
- **Glimmerfrost moves into the south hut** as its inhabitant; her charm is the visit reward, so reaching her before the raiders do is a real race.
- **The remaining two chwinga become Priests** on Sclorbo's class, with staves to heal themselves — survivable the way Ross and Garcia are, instead of fliers who could simply leave.
- The wiki confirmed vanilla's protect layer is Ross+Garcia and its villages are uncontested; ours makes one hut contested, which is the deliberate divergence.

## Dialogue

Boxes 1–3 of the south hut are vanilla's own `MSG_96C` **word for word** — it is the contested tile and vanilla already wrote the alarm that belongs on it. The east hut is an information house in vanilla's reward-free shape, paying in the one fact the chapter never states anywhere else: the chwinga repay you. Halvar's turn-1 bark announces the raid so the player has time to respond, rather than learning it by walking into a village.

`make check` green.

Refs #26, #335

### DIFF
--- a/campaigns/rime-of-the-frostmaiden/chapters/ch00-prologue-a-dagger-of-ice.yaml
+++ b/campaigns/rime-of-the-frostmaiden/chapters/ch00-prologue-a-dagger-of-ice.yaml
@@ -177,7 +177,14 @@ enemy_units:
       - "Cold regeneration, fire vulnerability — the Frostmaiden's chill knits his wounds; flame unmakes them (echoed later by the Ch8 ice trolls)"
       - "True nature (book p.23): a drowned mariner possessed by a frost-druid spirit of Auril; kill the body and the winter spirit dies with it — but a single defeat only scatters him"
       - "Misty step: vanishes in a breath of frost — the diegetic reason he escapes the prologue despite being 'defeated'"
-    ai_pattern: hold_position
+    donor: [14, 8]              # O'Neill, the tutorial boss, on this very tile
+    ai_override:
+      ai: '{GuardTileAI, 0x9, 0x20}'
+      why: >
+        O'Neill's own AI is DoNothing ({0x6,0x3,0,0}) and it only works because vanilla's
+        prologue event-scripts his attack -- prologue-eventscript.h CHAIs him to {0,0} once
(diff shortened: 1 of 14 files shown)


NicoviGH:
**Pre-merge review (fresh reviewer, not the author): clean.** No Critical, no Important.

Checked specifically for the failure mode worth worrying about -- that the difficulty
numbers were made to pass rather than earned. They were not: ch02's x0.81 threat / x0.79
clear-load is byte-identical before and after this PR, so the gap is a pre-existing property
of the roster and not something the decoy rework caused or quietly tolerates. Worth stating
in the description so a future session does not re-litigate it.

Also verified: the village/map-change wiring follows the established ch05 pattern rather
than inventing a per-chapter one-off; the new message ids sit inside ch02's own declared
block and are registered where the guard sees them; and the 13 lines dropped from
matrix.yaml are a documented migration of ch02's scenarios into its own YAML (per #314), not
lost coverage -- two scenarios were added on top.

Rebased onto the corrected #339 (95d8cb1). `difficulty-gate` exits 0 here, 793 tests pass.

PR #341, opened 2026-08-31
ch02: point slot 3 at its own map-change layer, so a hut can actually fall (#335)

NicoviGH:
**Stack 3 of 5.** Based on #340.

## The bug

`ch02raid` reported both Targos huts standing after twelve turns, which put the decoy #340 is built on in question. One instrumented run answered it: **the AI was never the problem.** Halvar walks onto the east hut on turn 1 with nothing within four tiles, decides `PILLAGE`, and stands there through turn 12 with `ai2` still `0x04` — re-deciding to loot a village that never falls.

The break is one line of ordering. `_retarget_host_chapter` zeroes the host slot's `map.changeLayerId`; `_inject_tile_changes` is what points it at the chapter's table. `inject_ch02` called them the other way round, so `MS_Ch02MapChanges` shipped and was registered while slot 3 kept pointing at `gChapterDataAssetTable[0]`.

Verified in the built data: slot 3 read `changeLayerId=0` where ch03/ch04/ch05 read 247/249/254 — ch02 was the only orphan.

Since `AiPillageAction -> StartAvailableTileEvent -> CallTileChangeEvent(GetMapChangeIdAt(...))` reads whatever layer the slot points at, **no tile change in ch02 could ever fire** — not sacking a hut, not closing a visited one. Nothing failed loudly; a village just stood through a raid, which reads as broken AI.

`check.py` now pins the order for every chapter injector, test-first.

## Also here

`inChapter()` conflated "the chapter this scenario is about" with "the chapter this boot is trying to reach" — identical for a direct boot, different for every chain. `ckpt_ch02start` replays ch00→ch01→ch02, and ch02's cases carry `host_chapter: 3`, so its first leg booted the prologue and waited for chapter 3 there. And `bootToMap` assumed New Game lands on the chapter's map, true only where a debug boot exists; ch02's checkpoint name was never exported to Lua at all.

**`ch02raid` PASSES:** targos-hut-east sacked on turn 2, terrain `0x03 -> 0x25`, no gift, flag unset.

`make check` green.

Refs #335, #26


NicoviGH:
**Pre-merge review (fresh reviewer, not the author).** The ch02 root-cause fix is correct
and the harness chain-vs-scenario disambiguation is sound and narrowly applied (`winCh00`
is the only site that needed it; `PROLOGUE = 1` checked against `hosts.py`).

But the regression guard this PR's description leans on -- "check.py now pins the injector
call order for every chapter" -- was false twice over. Both fixed in 7cf289b:

1. **It was never registered.** `check_tile_changes_outlive_the_retarget` was defined and
   unit-tested but absent from `main()`'s check tuple, while its sibling
   `check_cached_steps_are_config_invariant` -- added the same way for the same bug class --
   is there. It ran only as a side effect of `check_tests_pass` re-invoking the test file as
   a subprocess, and that no-ops whenever `fireemblem8u/src` is missing: exactly the
   lightweight CI job it was meant to hold.
2. **ch03 was invisible to it.** The guard matched the literal `_inject_tile_changes`, but
   ch03 calls `_inject_ch03_tile_changes`, a per-chapter wrapper. A skipped injector and a
   chapter with no map changes look identical from outside, so ch03 was ordered correctly by
   discipline rather than by construction. `check.py` had already solved this in
   `_injection_call_sequence` (which unwraps `_scopes.run`/`_anims.run` for the same reason,
   and says so).

Coverage is now assertable rather than implied: `_tile_change_injectors_seen` names who the
guard inspects and a test pins ch03 inside it. On the real `build_campaign.py` that set is
**ch02, ch03, ch04, ch05** -- up from three -- with ch01 correctly out (no tile changes) and
zero violations.

PR #342, opened 2026-08-31
playtest: a scene is not a stall, and an unfinished run is not a failure (#335)

NicoviGH:
**Stack 4 of 5.** Based on #341.

Five defects the ch02 work surfaced, all in shared harness code, all with the same shape: **a fixed budget standing in for "nothing is happening", on flows that legitimately contain a scene, end a chapter, or hold a unit off the map.**

## Budgets

`awaitControllerState` exempted only the event engine from its stall budget, so a legitimate **enemy phase** burned it — ch02's turn 1 runs seven raiders with map-animation battles for ~2100 frames and the wait gave up at 300, faulting a chapter it went on to rout. What counts as "the engine working" now lives in `Controller.engineIsWorking`, unit-tested against observations in `test_controller.lua` rather than asserted as a substring.

`runEnemyPhase` had two halves of one bug. ch04's parley phase plays a cutscene *and* the chapter ending inside the enemy phase: still running, chapter unchanged, 3,768 frames in, and the chapter not flipping until past 9,800. The flat 3,600 cap died mid-scene AND, once raised, the loop spun to the new cap because a chapter ending under it was not an exit. **It needed both** — each alone still failed.

## Coordinates

`struct Unit`'s `xPos`/`yPos` are `s8` and FE8 parks an off-map unit at −1, but `unitAt` read them with `ru8` — so such a unit reported **x=255**. The clear-bot took that for a real position and drove the cursor at (255,8) on a 25-wide map. ch02 printed the same number in `teleportToFiringTile: (255,9)` and only survived because that one site bailed.

The #220/#238 contract vets the *input* — is this press legal here — but nothing vetted the *destination* a caller asks for, so an off-map target walked the cursor to the wall and reported "cursor_left is not legal": true, and pointing at nothing.

And once ch01's boss dies the clear loop offers the seize tile to every remaining unit; the ones out of range pressed A on a tile FE8 correctly offers no confirm for. That sticky fault failed a run that **won by turn 9**. "This unit cannot reach that tile" is an answer, not a defect.

## The one that cost the most

`parse_verdict` read every non-PASS `RESULT:` line as FAIL — including run.sh's own `RESULT: ERROR -- timed out after Ns`. A `SUITE=all` sweep put six multi-minute scenarios in one concurrency block alongside `-j8` builds; they dropped to ~15fps against a 240fps target, timed out, and were tabled as **eight broken chapters**. All eight passed re-run serially. ERROR still blocks and is never cached — this changes what the table *says*, not what the gate lets through.

## Verification

`clear_ch01` went from stuck-on-turn-1 to **won by turn 9**. `clear_ch04_parley` and `clear_ch02` both green. Full `SUITE=all`: **50/50**. `make check` green.

Refs #335


NicoviGH:
**Pre-merge review (fresh reviewer, not the author).** The five described defects are each
fixed correctly. Verified in code rather than from the description: `awaitControllerState`
still has a hard outer bound (`total < budget * STALL_CEILING`) so the new "engine is
working" exemption cannot hang; and `RESULT: ERROR` genuinely still blocks -- `Report.failures`
counts `verdict != 'PASS'`, and both the store and load paths gate on `verdict == 'PASS'`, so
an ERROR is never cached as a pass.

**One Critical, fixed in dbbead6: the s8 change was applied to the two sites that
misbehaved and silently retired twenty-three others.**

Every other off-map test in the harness was written against the unsigned encoding
(`u.x ~= 0xFF`). `rs8` returns -128..127 and can never return 255, so all 21 of those became
vacuously TRUE and the two `== 0xFF` forms vacuously false. Nothing errored -- the guards
just stopped guarding, which is worse than the cursor crash the s8 read was fixing:

- five deployment counters counted units that are not on the map;
- `partyDeployed()` returned true for a unit that exists but has not been placed -- the exact
  false positive its own comment says it exists to prevent, and it gates the ch01->ch02 boot wait;
- the off-map lord check and the Baxby check could never fire again;
- two live-enemy counts stopped filtering off-map reds.

All 23 now read `.onMap`, the field `unitAt` already computes. `battleSide`'s diagnostic
decode of the same two fields moves to `rs8` with them.

Every one of the 23 variables was traced to `unitAt` (directly or via `findUnit`/`blue`/`red`,
all of which carry `.onMap`) before substituting, both inverted forms re-read in context,
24 lines changed and nothing else, Lua chunks load, top-level local headroom unmoved at 2.
The test pins the **encoding** (no `.x == 0xFF` / `.x ~= 0xFF` anywhere in the harness)
rather than the call sites -- pinning sites would need extending by whoever adds the
twenty-fourth, which is precisely what did not happen here.

**Behavioural proof is still owed on a matrix run** -- these are shared clear-bot helpers,
so the honest tier is `SUITE=all`, and per #345 it must run `--jobs 1` or it will report
false failures.

PR #343, opened 2026-08-31
docs + tooling: one guidance file, AI visible per chapter, and a message-id guard that stops trusting names (#335)

### LINKED ISSUE #335: AI is authored, not derived — audit ch00–ch05 against their vanilla twins, then guard it

NicoviGH:
**Stack 5 of 5.** Based on #342. Closes #335.

## 1. AGENTS.md and CLAUDE.md stop duplicating each other

Both described how to work in this repo — the drift this project guards against. `AGENTS.md` had silently fallen behind: same fourteen headings, twenty fewer lines, and a base-ROM path still pointing at `Documents/Codex`, so any agent reading it got wrong instructions.

Per [Anthropic's guidance](https://code.claude.com/docs/en/memory): Claude Code reads `CLAUDE.md`, not `AGENTS.md`, so a repo keeping `AGENTS.md` for other agents should have a `CLAUDE.md` that **imports** it. `AGENTS.md` now carries every tool-agnostic instruction; `CLAUDE.md` is `@AGENTS.md` plus the one genuinely Claude-specific section. `check.py`'s comment-drift guard scans both, or the content would have walked out from under it.

## 2. `make chapter CH=chNN` shows where each enemy's AI came from

#335's third scope item. Per enemy: the behaviour family its AI2 script belongs to (pursue / never-move / pillage / charge-after-1 / guard-tile, read off `gAi2ScriptTable`) and the donor coordinate it was borrowed from — or `override` plus its stated reason. An unresolvable donor prints `UNGROUNDED` rather than dropping the unit off the report.

## 3. The message-block guard stops trusting a naming convention

Found by reading the merged #338 diff, then confirmed by the review subagent finding **two more of the same class**:

- `injector_message_ids` located ids by scanning globals whose **name** ends in `_MSG`/`_MSGS`, while promising "registering a new one is enough — there is no second list to remember". False for ids held in a constant named anything else (ch02's hut visits in `CH02_VILLAGE_SLOTS`).
- `PC_DEATH_QUOTE_MSGS` is a **dict** whose name *does* match — the flattener wrapped it as `(the_dict,)` and all 13 death quotes vanished.
- An undocumented `0x300` floor exempted every chapter's objective window and status string, and applied to the claims fold too, so registering them did not help.
- The prologue and ch01 wrote **12 bare literals** no source could see. `0xC25` was the sharp one: `0x33` above ch05's pool, so extending that range upward — the obvious next move — would have been accepted and overwritten a defeat quote.

Each widening surfaced the next problem. The last one found a **real overlap**: ch04's block ran to `0x9CC` while ch04 spends only through `0x9C6`, and ch05 writes `0x9C9`/`0x9CA` inside that tail; ch05 was also spending `0x9CD-0x9D2` with no block declaring them, so `make chapter` read its headroom wrong. ch04 narrows to `0x9BA-0x9C8`, ch05 declares the range it was already writing. **No shipped id moves** — this makes the declarations true.

Also: the guard ran only from the tests, so `make check` caught a bad block but a plain `make` still produced a ROM. It now runs in `main()` beside its two siblings. `HOSTED_VANILLA_SLOTS` was read by nothing while its comment described the safety property as enforced — removed.

## On #335's scope

Item 2 asked for a chapter-level `ai_divergence:` allowlist. It is **deliberately not built**, and the ADR says why: a donor carries its AI, so there is no authored second opinion for an allowlist to referee, and per-unit `ai_override:` with a mandatory `why` is the right granularity — a donor is chosen per unit. The stretch (AI in the difficulty metric) is filed separately.

`make check` green. Full `SUITE=all`: 50/50.

Closes #335


NicoviGH:
**Pre-merge review (fresh reviewer, not the author): no correctness defects.** Verified
directly: the ch04/ch05 overlap is actually FIXED here, not merely detected
(`live_ids_in_declared_blocks()` returns empty against the real blocks); dict-valued and
sub-`0x300` ids are genuinely caught now; the build fails loudly on a real collision where
before this check only ran from tests; and CLAUDE.md/AGENTS.md content parity is intact with
nothing dropped in the move.

Two adjustments:

1. **Retitled.** "a guard that sees every id" oversold it. The bare-literal case is a
   hand-curated allowlist, not a scan -- the next bare literal in a new chapter's injector is
   still invisible until a human notices. The docstring is honest about this; the title was
   not. Filed as #346 rather than built here, to keep this PR landable.
2. **`AiRow.override` stored the AI vector while two call sites test it for truth** (6639431).
   Harmless today since every vector is a non-empty string, but an override whose vector ever
   stringified to something falsy would drop out of both the donor-column label and the
   printed reason while still being an override. It stores the flag now.

**One review finding rejected.** The reviewer read the commit message's "AGENTS.md had
silently fallen behind, still pointing the base ROM at Documents/Codex" as fabricated,
because git shows `AGENTS.md` first added in this very commit. That inference does not hold:
`AGENTS.md` was **untracked** (HANDOFF.md says so explicitly), so git cannot testify to its
prior content, and `Documents/Codex` is a real path in this repo's history -- it is in
HANDOFF.md at 383c047 as the old checkout root. The shipped path in the new file is the
correct `Documents/Claude/...`. No change made.

PR #356, opened 2026-09-02
tooling: a bare-literal message id registers itself (#346)

### LINKED ISSUE #346: message-id discovery: catch a bare literal without a human registering it
Follow-up from the review of #343 (deliberately kept out of that PR to keep it landable).

#343 widened `injector_message_ids` so it sees ids held in dicts, in constants named
anything, and below the old undocumented `0x300` floor. That is a real improvement, and it
found a real overlap: ch04's declared block ran past what ch04 spends, into ids ch05 writes.

**What it does not do is scan for bare literals.** The prologue and ch01 write 12 message
ids as bare hex in the injector source. Those are caught today only because someone grepped
for them and hand-transcribed them into `PROLOGUE_LITERAL_MSGS` / `CH01_LITERAL_MSGS` --
tuples named to match the `_MSGS?$` convention so the existing discovery picks them up.

So for that class the guard's promise ("registering a new one is enough -- there is no
second list to remember") is still false: the next bare literal someone writes into a new
chapter's injector is invisible until a human notices it. The docstring is honest about
this; the mechanism is what is missing.

`0xC25` is the sharp example of why it matters: it sits `0x33` above ch05's pool, so
extending that range upward -- the obvious next move -- would have silently overwritten a
defeat quote.

**Suggested shape:** a narrow source scan over `set_message_body(lines, 0x...)` call sites
in `build_campaign.py`, folded into `injector_message_ids` or a sibling, so a bare literal
registers itself. Cheap, and it closes the last naming-convention dependency.

Found by the pre-merge review of the #335 stack.

NicoviGH:
Closes #346.

## The problem

`injector_message_ids` finds a message id by the **name of the constant** holding it, and its
comment promised that "registering a new one is enough — there is no second list to remember".
That was false for an id passed as hex at the `set_message_body` call site. Twelve such bare
literals exist on `main` (8 prologue + 4 ch01) and they reached the deadness guard only because
somebody grepped for them once and hand-transcribed them into `PROLOGUE_LITERAL_MSGS` /
`CH01_LITERAL_MSGS`. The thirteenth would have been invisible.

**`0xC25` is the sharp case.** It sits `0x33` above ch05's `0xBC5-0xBF2` pool, so extending that
range upward — the obvious next move when ch05 runs out of ids, and `make chapter CH=ch05` already
reports that block as FULL — would have been accepted by every guard in the repo and would have
overwritten Scramsax's defeat quote.

## The fix — two halves, because #346 is two problems

**Discovery.** `inject.hosts.literal_message_ids` AST-scans `build_campaign.py`'s own source
through `tools/callsites.py` for `set_message_body` call sites with an integer-literal `msg_id`,
and attributes each to the injector that writes it (`inject_prologue` → ch00, `inject_chNN` →
chNN). It reads argument **bindings**, not text: `msg_id` is positional at all 71 call sites, so
`grep msg_id=` finds exactly none of them. `injector_message_ids` folds the result in, so the
build-time gate and `make chapter` headroom see a bare literal with **no human step at all**.

**Ownership.** Discovery cannot invent an owner, and ownership is what `assert_message_ids_unique`
collides on and what `make chapter` reads for headroom — an id spent but unclaimed reads as free
room that has already been spent. `check_message_literals_are_registered` still requires the claim
and names the tuple to add it to.

Split cleanly: **discovery makes the id safe, registration makes it accountable.**

## Anti-#341 measures

Per the 2026-09-02 ADR *"A guard that stops guarding fails silently, and green is what that looks
like"*:

- registered in `check.py`'s `main()` tuple, **pinned by a test** that reads
  `inspect.getsource(check.main)` — a check that is defined and unit-tested is still not *running*;
- **vacuity guard**: a live scan finding zero literals *fails*, because a broken scan and a clean
  tree are the same silence;
- `literal_message_ids` raises rather than returning `()` if the writer's signature can't be
  resolved, since an unresolvable signature silently disables positional binding in `callsites.scan`.

Reuse over new machinery: `callsites.scan` already does argument binding (same reason
`check_wrap_widths_are_pixels` reads bindings), and `inject/hosts.py` already source-scans
`build_campaign.py`, owns `_INJECTOR_RE`, and is stdlib-only — so the lean CI `checks` job runs this
guard for real instead of skipping it.

## Verification

| | |
|---|---|
| `python3 tools/test_check_message_literals.py` (18 tests) | **before: 17 errors + 1 failure · after: 18 pass** |
| `python3 tools/check.py` | `drift check: clean`, exit 0 |
| Bare literals discovered on `main` | **12, all already claimed** — the guard ships **strict, no allowlist** |
| Not vacuous | planting `set_message_body(lines, 0xBF3, …)` into `inject_ch05` produces the full complaint naming `CH05_LITERAL_MSGS` |

**No ROM build and no playtest are needed** — this is discovery plus a lint. Nothing here touches
injected data, the decomp, or anything the ROM contains.

## Review

Per the 2026-09-02 ADR *"`/code-review` is the review step, and the author never reviews their own
PR"* — **this has not been reviewed yet.** It was drafted by a subagent and independently verified by
me (I ran the before/after myself rather than taking the agent's word), but that is not a review.
Needs a fresh reviewer before merge.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4237,15 +4237,26 @@ BINDINGS, not text, because `msg_id` is positional at all 71 call sites and `gre
 none of them -- and attributes each to the injector that writes it. `injector_message_ids` folds
 them in, so deadness is checked with no human step.
 
-**Discovery cannot invent an OWNER, so the registry stays, and it stays a gate.**
-`HOSTED_CHAPTER_MESSAGE_IDS` is what `assert_message_ids_unique` collides on and what
-`make chapter` reads for headroom; an id spent but unclaimed reads as free room already spent.
-`check_message_literals_are_registered` requires the claim and names the tuple to add it to. The
-two halves split cleanly: **discovery makes the id safe, registration makes it accountable.**
+**Discovery cannot invent an OWNER, so the registry stays -- but the ownership check belongs
+where the registry is REAL.** `HOSTED_CHAPTER_MESSAGE_IDS` is what `assert_message_ids_unique`
+collides on and what `make chapter` reads for headroom; an id spent but unclaimed reads as free
+room already spent. The first cut asserted that in `check.py` by reading the registry with an AST
+evaluator, and #356's review killed it: the registry is written as generators, subscripts and
+tuple-unpacked constants (`*(msg for (_slot, msg, _boxes, _what) in CH05_OPENING_SLOTS)`,
+`CH05_ARRIVAL_SLOT[1]`, `CH04_VILLAGE_MSG`), so the static read was wrong in **both** directions
+-- it missed ch04's `0x9C3`/`0x9C6` and would have demanded a registration that makes
+`assert_message_ids_unique` exit, and it swept up ch05's box counts so an unclaimed `0x13` passed.
+**A hand-rolled evaluator of a data structure is a second, worse implementation of it.** Ownership
+now lives in `build_campaign.assert_literals_are_claimed`, called from `main()` beside
+`assert_message_ids_unique`, where the dict is a real Python object and the answer is exact.
 
-The guard is registered in `check.py`'s `main()` tuple and pinned by a test that reads that
-function's source, and it fails when the live scan finds zero literals -- both because of the
-lesson directly above.
+The split that survives: **`check.py` owns DISCOVERY** -- stdlib-only, so the lean CI `checks` job
+runs it for real -- and **the BUILD owns OWNERSHIP.** Discovery makes the id safe (it folds into
+`injector_message_ids`, so deadness is checked with no human step); ownership makes it accountable.
+
+The discovery guard is registered in `check.py`'s `main()` tuple and pinned by a test that reads
+that function's source, and it fails when the live scan finds zero literals -- both because of the
+lesson directly above. The build-time half is pinned the same way, by a test reading `main`.
 
 - **Dressing a portrait slot and NORMALIZING its mouth/eye window are two steps, and missing the
   second is silent** (2026-08-09, #25). `patch_portrait_geometry` only knew about `PORTRAIT_MAP`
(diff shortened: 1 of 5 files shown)


NicoviGH:
## Review findings addressed — the ownership half was a design error, not four bugs

All four findings were real. Fixing them one by one would have kept a hand-rolled AST evaluator of `HOSTED_CHAPTER_MESSAGE_IDS`, and that registry is written as generators, subscripts and tuple-unpacked constants — so it would have kept being wrong. **Ownership moved to build time, where the dict is a real Python object.**

| Finding | Resolution |
|---|---|
| `check.py:2017` unresolvable claim → unsatisfiable failure | **Gone with the static read.** Verified with your reproduction: `assert_literals_are_claimed(literals=[ML(0x9C3,'ch04',1)])` now passes — ch04's tuple-unpacked claims resolve exactly, because they are evaluated, not parsed |
| `check.py:1905` over-collection → unregistered literals pass | **Gone with the static read.** Verified: a planted `0x13` in ch05 now exits the build instead of producing zero findings |
| `check.py:1980` `ParseError` escapes and kills the gate | Caught and reported; new test `test_a_malformed_source_is_REPORTED_not_raised` pins it |
| `test…:132` hardcoded count of 12 | Replaced with count-free assertions (non-empty, all attributed, `0xC25` present), per the repo's own convention |
| PR body overstated `make chapter` headroom | Corrected below |

**The split that survives:** `check.py` owns **discovery** (stdlib-only, so the lean CI `checks` job runs it for real); `build_campaign.assert_literals_are_claimed` owns **ownership**, called from `main()` beside `assert_message_ids_unique`. Discovery makes the id safe; ownership makes it accountable.

The ADR in `docs/decisions.md` was rewritten to describe the design that survived review rather than the one that shipped in the first commit, and it now records *why* the static read failed — a hand-rolled evaluator of a data structure is a second, worse implementation of it.

**Correction to the PR body above:** `make chapter` headroom reads `HOSTED_CHAPTER_MESSAGE_IDS` only, not `injector_message_ids`, so "the build-time gate **and `make chapter` headroom** see a bare literal with no human step" was overstated. Discovery reaches the deadness check; headroom still reads the registry, which is exactly why the ownership half has to exist.

Verification: **21 tests pass**, `python3 tools/check.py` clean (exit 0), both of your reproductions confirmed fixed. Still no ROM build or playtest needed.

Issue #347, opened 2026-09-02
ch01's goblins ship as ch05's skeletons: a reskin is resolved by BASE, and base is many-to-one


NicoviGH:
**ch01's goblins ship as skeletons.** Found by Nicolas playing; confirmed in the emitted
data, not off the screen.

`fireemblem8u/src/events_udefs.c` gives ch01's grunts `CLASS_SOL_SKELEBERDIER` and
`CLASS_FGT_SKELEBERDIER` — ch05's risen tomb-guard — where the campaign declares the Fire
Imp goblin reskin.

## Cause

`build_campaign.py:7954`, inside `inject_ch01`:

```python
reskin_by_base = {rk['base']: rk['slot'] for rk in enemy_class_reskins(campaign)}
```

A dict keyed on the **base** class, so last-wins. Four bases are claimed twice:

| base | first claim | later claim (wins) |
|---|---|---|
| `CLASS_SOLDIER` | goblin-soldier → fire-imp | risen-spear → skel-lance |
| `CLASS_FIGHTER` | goblin-fighter → fire-imp | tomb-reaver → skel-axe |
| `CLASS_BRIGAND` | kobold-grunt | kobold-brute |
| `CLASS_MERCENARY` | kobold-blade | crypt-blade |

**`base` is many-to-one by design** — a reskin CLONES a vanilla class into its own slot, and
several creatures legitimately clone the same chassis. Slots are unique (verified); bases
are not. So base can never be a resolution key, and the duplicates are not the defect.

## Scope

**ch01 only.** ch03/ch04/ch05 name their slots explicitly in `CHnn_CLASS_IDS` — ch05 has
`'soldier': 'CLASS_SOL_SKELEBERDIER'` — so they resolve correctly by construction. ch01 is
the sole chapter that names vanilla classes and remaps through `reskin_by_base`. ch02 names
vanilla classes and does NOT remap, which is correct (its raiders are human).

The chief is a separate, smaller thing: `armor-knight` has no reskin, so he stays a vanilla
Knight among goblins.

## Introduced

`e7d1887` (2026-08-21), "ch05: dress the risen tomb-guard as skeletons". Adding ch05's
skeletons silently stole ch01's goblins. Twelve days undetected: nothing asserts which
sprite a unit wears. #48 measures stats, the playtest scenarios assert memory state, and
neither reads identity — the same blind spot as #335, one layer over.

## Why this blocks ch06

[...]

Issue #42, opened 2026-06-08
Player-chosen 'lord' (must-survive lead) before Ch1


NicoviGH:
**Idea (Nicolas):** before Ch1, the player chooses one PC to be their "lord" — presumably the character they actually played in the D&D campaign. That unit then carries the **lord-death lose condition** for the rest of the game (personalized stakes).

**Feasibility (traced in the decomp):** game-over fires in `bmtrick.c` when `CheckFlag(EVFLAG_GAMEOVER) || CountAvailableBlueUnits()==0`. Today `EVFLAG_GAMEOVER` is set per-character via the death-quote table (`data_battlequotes.c`) — i.e. it's character-driven, not hard-wired. So this is clean to make player-selectable.

**Scope (generic engine mechanic — belongs in `engine/`, no character names in C):**
- [ ] Pre-Ch1 **selection menu** — list the 7 PCs, pick one (the bulk of the work: UI).
- [ ] **Persist** the choice in a save slot.
- [ ] **Hook unit death** to set `EVFLAG_GAMEOVER` when the *chosen* unit dies.
- [ ] (optional) grant the chosen unit `CA_LORD` so they can Seize / lead.
- [ ] Default/fallback when unset (e.g. a designated lead) so chapters never soft-lock.

**Sequencing:** not needed for the Ch0 slice (Ch0's lead is Hlin; the party doesn't exist yet). Build before Ch1 ships. Until then, chapters can designate a must-survive lead via YAML/events.

Refs: ch00 YAML lose-condition NOTE 2; #20.

PR #357, opened 2026-09-03
playtest: ch01 gets a debug boot, and a boot must strip MENUS too (#353)

### LINKED ISSUE #353: ch01 has no debug boot: confirming anything there costs a prologue playthrough
**ch01 is the only hosted chapter with no debug boot.** The ROM configs are `testch`,
`ch03boot`, `ch04boot`, `ch05boot`, `ch05lupinboot` — ch03 through ch05 can be reached in
seconds, and ch01 can only be reached by playing the prologue.

`scenarios.ch01` calls `reachCh01Map()`, which wins ch00 first. Confirming #347 visually cost
a full canonical build plus a prologue playthrough (~24,000 frames) to look at a map that
ch03 would have shown in about twenty seconds.

That is the exact case `decisions.md` already argues for: build the fast boot FIRST so
iteration is compile-time only, rather than replaying to reach a late screen.

**Why it is worth it now rather than later.** ch01 is where the goblins live, the reskin work
is active (#347 just moved their classes), and it is the first real chapter a playtester
sees — so it gets looked at more than its lack of a boot suggests. Every one of those looks
currently costs a prologue.

**Shape:** the same as the four that exist — a `CH01BOOT=1` flag that configures New Game to
land on ch01's map with the party deployed, plus its scenarios retargeted to that config.
`inject_ch01`'s existing host retarget already does most of the work; `_configure_boot` is the
pattern (`CH05_HOST_INDEX` and friends).

Found while confirming #347 by eye, which is what surfaced the cost.

NicoviGH:
Closes #353.

ch01 was the only hosted chapter without a fast boot, so confirming #347 by eye cost a canonical
build plus a full prologue playthrough (~24,000 frames) to see a map ch03 would have shown in
twenty seconds.

## What made ch01 different

**No seed table.** ch01 is the chapter that *founds* the party — its opening LOADs the company at
the Northlook and runs PREP — where `--ch03/04/05-boot` must conjure an armed party from nothing
(`CH03_BOOT_SEED_SYMBOL` and friends). The boot reuses ch01's own LOAD, so a cold start founds the
party exactly as the canonical chapter does.

**Lord select (#42), which nothing predicted.** The first cut kept ch01's opening and the run died
with `fail:nil ... never reached the map` on an 8-item `generic_menu` — the lord-select menu. No
other boot has met one, because no other chapter opens with a menu. Hence the lesson in the ADR:
**a debug boot is a MAP load-test, so it must strip every screen between New Game and the map, and
a MENU is such a screen even though it is not a cutscene.**

## Measured, not reasoned

| | |
|---|---|
| New Game → `player_map_idle`, chapter 2, player faction, turn 1 | **frame 2,856** (vs ~24,000 for the honest route) |
| `PT_HOST_CHAPTER=2 tools/playtest/run.sh mapshot` | **PASS** — "map deployed; screenshots taken" |
| Same via `matrix.py` | FAIL — `mapshot` is chapter-generic (`host_chapter: null`) so it resolves to host 1 while the ROM boots chapter 2, waits for a map that never matches, reports `boot-timeout`. Scenario wiring, not a boot fault; documented so it is not rediscovered |

## Two guards, both written RED first

- **`check_rom_configs_reach_the_build`** — a `rom_config` env var the Makefile does not translate
  builds **canonical**, so the scenario's verdict is about a different ROM than the one it names.
  Silent in both directions. Adding `ch01boot` without the Makefile arm fires it; adding the arm
  clears it.
- **`check_decomp_git_calls_strip_the_env`** — see below.

## The blocker this had to clear first

**The pre-commit hook could not pass from a worktree at all.** `git commit` exports
`GIT_DIR`/`GIT_INDEX_FILE` to hooks, and an explicit `GIT_DIR` **beats `git -C`** — so every
`git -C fireemblem8u show HEAD:...` resolved against the superproject and exited 128. It surfaced
as 12 errors in `test_build_campaign.py` that appear **only** under `git commit` and never when you
run the tests yourself, which is exactly why it read as "my change broke the tests".

`vanilla_decomp_text` already knew this — its comment says *"Bit us committing from a
content/pipeline worktree"* — but the lesson lived in **three inline copies** of the same env-strip
dict, and the one site that mattered (`chapter_label_constants`, the same `show HEAD:` call) had
none. It is now `inject.decomp.git_env()`, used at all 8 `git -C DECOMP` call sites and pinned by
AST.

**Proof:** with `GIT_DIR`/`GIT_INDEX_FILE` set exactly as the hook sets them,
`test_build_campaign.py` went from `FAILED (errors=12)` to **607 tests OK** — and this PR's own
commit passed the hook.

## Deliberately not done

Existing ch01 scenarios are **not** retargeted. `ch01win` and friends prove the ch00 → ch01 chain
works; repointing them at a boot that skips the prologue would delete exactly that coverage — the
*"a scenario written against the old design will FAIL ON SUCCESS"* trap. The repo's own precedent is
a fast **twin** (`recordlord` / `recordlordfast`), not a retarget. A dedicated fast ch01 record
scenario would cost one of `harness.lua`'s **2 remaining** top-level local slots, so it is noted on
the issue rather than spent here.

## Review

Not reviewed yet — needs `/code-review` before merge, per the ADR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ all: fireemblem8.gba
 # MONTAGE=1 wires the #43 opening montage (lore crawl on New Game) in place of
 # the dev straight-to-map boot cut. Distribution builds (#37) must set it.
 fireemblem8.gba:
-	python3 tools/build_campaign.py --campaign $(CAMPAIGN) $(if $(MONTAGE),--montage) $(if $(TESTCH),--test-chapter) $(if $(LORDBOOT),--lord-boot) $(if $(CH03BOOT),--ch03-boot) $(if $(CH04BOOT),--ch04-boot) $(if $(CH05BOOT),--ch05-boot) $(if $(CH05LUPIN),--ch05-lupin) $(if $(CH05MOOSE),--ch05-moose) $(if $(CH05ENDING),--ch05-ending=$(CH05ENDING))
+	python3 tools/build_campaign.py --campaign $(CAMPAIGN) $(if $(MONTAGE),--montage) $(if $(TESTCH),--test-chapter) $(if $(LORDBOOT),--lord-boot) $(if $(CH01BOOT),--ch01-boot) $(if $(CH03BOOT),--ch03-boot) $(if $(CH04BOOT),--ch04-boot) $(if $(CH05BOOT),--ch05-boot) $(if $(CH05LUPIN),--ch05-lupin) $(if $(CH05MOOSE),--ch05-moose) $(if $(CH05ENDING),--ch05-ending=$(CH05ENDING))
 	$(MAKE) -C fireemblem8u fireemblem8.gba -j$(NPROC)
 
 # Drift guard: docs/tooling consistency. Same logic CI and the git pre-commit hook run.
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4221,6 +4221,35 @@ The unifying test: after any change to a check, an encoding, or a sentinel, ask 
 now be different if this were broken?* If the answer is "nothing observable", that is the
 bug, not the reassurance.
 
+**ch01 is the only hosted chapter whose opening runs a MENU, and that is what a debug boot has
+to strip (2026-09-02, #353)**
+Every hosted chapter but ch01 had a fast boot; confirming #347 by eye therefore cost a canonical
+build plus a full prologue playthrough (~24,000 frames) to look at a map ch03 would have shown in
+twenty seconds. The boot itself was the easy half -- `_configure_boot(CH01_HOST_INDEX)`, the same
+call ch03/ch04/ch05 make.
+
+**Two things made ch01 different, and only one was predictable.** ch01 needs **no armed seed
+table**: it is the chapter that FOUNDS the party, so its own opening LOADs the company at the
+Northlook and runs PREP, where ch03/ch04/ch05 boot into a party that must be conjured from nothing
+(`CH03_BOOT_SEED_SYMBOL` and friends). The unpredictable half was **lord select (#42)**. The first
+cut kept ch01's opening intact and the run died with `fail:nil ... never reached the map` on an
+8-item `generic_menu` -- the lord-select menu, which no other boot has ever met because no other
+chapter opens with one. **A debug boot is a MAP load-test: it must strip every screen between New
+Game and the map, and a MENU is such a screen even though it is not a cutscene.** The boot branch
+drops the Northlook beats, lord select and Preparations, and LOMAs straight to the trail.
+
+Measured: New Game -> `player_map_idle` on chapter 2, player faction, turn 1, at **frame 2,856**
+against ~24,000 for the honest route. `PT_HOST_CHAPTER=2 tools/playtest/run.sh mapshot` PASSes;
+going through `matrix.py` instead does not, because `mapshot` is chapter-generic
+(`host_chapter: null`) and resolves to host 1 while the ROM boots chapter 2 -- the run then waits
+for a map that never matches and reports `boot-timeout`. That is scenario wiring, not a boot fault,
+and it is why the invocation is documented rather than left to be rediscovered.
+
+Also from this change: **a rom_config env var the Makefile does not translate builds CANONICAL**,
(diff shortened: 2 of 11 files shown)


NicoviGH:
## All five findings addressed

All four bugs were real, and the fifth was a docstring claiming more than the code did. Rather than patch three lists by hand, the guard now covers them — this was the same silent-wrong-ROM class it was written for, so it should have from the start.

| Finding | Resolution |
|---|---|
| `CH01BOOT` missing from `_requested_flags` | Added. `check_rom_configs_reach_the_build` now **also** checks the build stamp |
| Guard scanned three hardcoded files | Now walks **every** `.py` under `tools/`, and additionally rejects `env=os.environ` (looks like the fix, is the bug) and any file rolling its **own** inline strip |
| `--ch01-boot` missing from `_boots` | Added; verified — `CH01BOOT=1 CH03BOOT=1` now exits naming both flags |
| `FLAG_ARGS` has no `CH01BOOT` | Added, plus the pre-existing `CH05MOOSE`/`CH05ENDING` gaps you didn't ask about but which would KeyError the same way |
| rom_configs regex `(?=^\w)` fragile at EOF | Now `(?=^\S\|\Z)` |

**On `gen_chapter_title.py`:** it does pass `env=`, so it wasn't unguarded — but you were right that it mattered. Its inline strip was missing `GIT_COMMON_DIR`, `GIT_NAMESPACE` and `GIT_ALTERNATE_OBJECT_DIRECTORIES`, so it **half-worked**. There were five inline copies of that dict in the repo; there is now one `git_env()`, and the guard rejects a sixth.

**One finding of my own, worth more than the rest.** While fixing these I noticed a test asserting on `inspect.getsource` failing impossibly — the widened git-env guard was sitting **dead behind a stale duplicate definition** of itself. Python keeps the LAST `def` and says nothing. A sloppy source-slicing edit did this **twice in one session** to two different guards in `check.py`; both times the module imported cleanly and the tests passed while the edits did nothing. New guard `check_no_shadowed_definitions` rejects duplicate top-level defs across `tools/`. Same family as *"A guard that stops guarding fails silently"*, and now an ADR.

Also noted, no change made: the `ch01boot` config is referenced by no scenario and can't be driven through `matrix.py` today (documented in the ADR). A dedicated fast ch01 record scenario would cost one of `harness.lua`'s **2 remaining** top-level local slots, so that decision is on #353 rather than spent here.

**Verification:** 193 tests pass, `tools/check.py` clean, and the boot itself is unchanged since the in-engine measurement (frame 2,856 to `player_map_idle` on chapter 2).

NicoviGH:
## Round 2 addressed — all four, including two of my own guards failing at their own job

| Finding | Resolution |
|---|---|
| `CH05ENDING` mapped to a bare `--ch05-ending` | `VALUED_FLAGS` now carries the arm → `--ch05-ending=full`. You were right that the guard was *certifying coverage the consumer didn't have* |
| Three registry arms gated on `if <text> and …` | A region that can't be found is now a **finding**, not a skip |
| `rolls_own = 'git_env' not in text` was file-scoped | Guard now demands the `env=` expression **be** `git_env()` (or a local bound to it) |
| Unguarded `ast.parse` | Reported, not raised |

**The third one was the worst and you called it exactly.** In `build_campaign.py` — which mentions `git_env` — every *new* `git -C DECOMP` call would have passed, and `env=os.environ.copy()` slipped both tests at once: the file-scoped mention said "safe", and it is a `Call`, so the `os.environ` Attribute check never saw it. The guard would have reported clean on the precise #353 failure it was written for.

Verified against all four shapes now:

| Shape | Verdict |
|---|---|
| `env=git_env()` | passes |
| a local bound to `git_env()` | passes |
| `env=os.environ.copy()` | **rejected** |
| a private inline env dict | **rejected** |
| a file that does not parse | **reported**, not raised |

Each behaviour is pinned by a test that **plants the offending source and asserts the finding**, rather than asserting against the live tree — so these cannot pass vacuously once the tree is clean.

**201 tests pass, `tools/check.py` clean.** The boot itself is untouched since the in-engine measurement (frame 2,856 to `player_map_idle` on chapter 2).

Ready to merge unless you see more.

PR #358, opened 2026-09-03
playtest: the deadline is a FRAME budget, because a wall clock measures the machine (#345)

### LINKED ISSUE #345: SUITE=all reports false failures: four long scenarios at once starve each other past their wall-clock deadlines
## What happened

A `SUITE=all` sweep tabled **eight chapters as FAIL**. All eight passed when re-run with `--jobs 1`. The runs had not failed — they had been killed on run.sh's wall-clock deadline while the machine was contended, and `parse_verdict` was reading `RESULT: ERROR -- timed out after Ns` as FAIL.

The verdict conflation is fixed (see the #335 stack). **The contention that caused the timeouts is not.**

## The underlying problem

`resolve_jobs` defaults to `cpus // 2` capped at `MEASURED_JOBS = 4`, measured in #310 as "3.2x with no per-scenario slowdown". That measurement was taken on the **gate** suite, which is mostly short scenarios.

`SUITE=all` is a different shape. The `canonical` block holds nearly every long scenario in the repo — `smoke_ch01`, `clear_ch01`, `fuzz_ch01`, `ch01win`, `lordfloor`, `ch01lord`, `recordsupply`, `clear` — so four multi-minute mGBA processes run concurrently, interleaved with `-j8` ROM builds. Measured throughput in the bad sweep: roughly **15fps against a 240fps target**, i.e. ~16x slower than a solo run. Wall times of 46-58 minutes for scenarios that take 25-70 seconds alone.

Every deadline in `matrix.yaml` is wall-clock, so the slowdown converts directly into false timeouts, and it hits precisely the tier meant to be the pre-playtest safety net.

## Options (not yet decided)

1. **Make the deadline frame-based rather than wall-clock.** The work is frame-bound; the deadline is not. This is the honest fix but touches run.sh's kill path.
2. **Budget concurrency by expected cost.** The manifest already knows each scenario's `deadline`; treat that as a cost estimate and cap concurrent long scenarios (e.g. at most one scenario over N seconds at a time) rather than capping the count flatly.
3. **Report throughput.** The harness stamps every log line with its frame. Comparing final frame against wall time gives measured fps for free; a run finishing far below its target rate should say so in the table, whatever the verdict.

(3) is worth doing regardless — it is what makes a contention failure legible instead of a mystery. (1) or (2) is the actual fix.

## Repro

```
make matrix SUITE=all           # on a machine also doing other work
# vs
python3 tools/playtest/matrix.py run --jobs 1 --scenarios smoke_ch01,clear_ch01,fuzz_ch01,ch01win,lordfloor,ch01lord,recordsupply,clear_ch02
```

The second passed 8/8 in 4m07s.

NicoviGH:
Closes #345.

A `SUITE=all` sweep tabled **eight chapters as FAIL**. Nothing had failed. That tier puts nearly
every long scenario in the repo in one `canonical` block, so four multi-minute mGBA processes ran
against `-j8` ROM builds at roughly **15fps against a 240fps target** — 16× slower than solo — and
every deadline in `matrix.yaml` was **wall-clock**. Scenarios that pass alone in 25–70s blew
300–600s walls. All eight passed under `--jobs 1`.

## The fix

**The work is frame-bound, so the budget is frames.** `deadline` is now read as what it always
meant — how much *work* a scenario may do — as `deadline * fps` frames at its own target rate.
Contention changes how long a run takes and nothing about its verdict. On an idle machine the
arithmetic is identical, so the healthy case did not move (`mapshot`: 100,800-frame budget, real run
2,856).

**A frame budget alone would hang forever** on a wedged emulator, which emits no frames and so never
reaches any budget. Two kill paths, both naming a real fault rather than a busy laptop:

| | |
|---|---|
| **OVERRUN** | more frames than budgeted — genuinely not finishing |
| **STALL** | the frame counter stopped — mGBA is wedged |

`PT_MAX_WALL_S` remains a last-resort net for neither. Both new knobs are registered in
`PLAYTEST_ENV_KEYS` — the repo's own guard caught that omission, and it matters: a run killed early
by a tightened `PT_STALL_S` must not be served later as a cached PASS.

**Every run now prints `throughput: N frames in Ns = Nfps (target Nfps)`.** The harness already
stamped every line with its frame, so measured fps was free the whole time; printing it is what makes
a contended run legible instead of a mystery. The fix and the diagnosis are the same number.

## Verified end-to-end on `titlecard`, all three paths

| Path | Result |
|---|---|
| healthy | `5267 frames in 6s = 877fps (target 240fps)`, PASS, exit 0 |
| OVERRUN (1s budget) | `OVERRAN its frame budget (3803 > 240 frames, i.e. 1s of work at 240fps)`, exit 1 |
| STALL (`PT_STALL_S=0`) | `STALLED: no frame progress for 0s (stuck at frame 0)`, exit 1 |

`matrix.yaml` was restored after the budget test; the diff carries no scenario change.

## Two bugs found while proving it

- **`set -euo pipefail` turns an empty `grep` into a dead script.** Before the first stamped line
  exists, the frame scan finds nothing, `grep` exits 1, and `run.sh` died **silently** right after its
  "running" banner — while the scenario went on to PASS in its own log. No frames yet is a normal
  early state, not an error.
- **Bash reads a leading-zero number as OCTAL in `$(( ))`.** Stamps are zero-padded, so frame 3425
  evaluated as 3425₈ = 1813 and the throughput line under-reported by a third. `10#` forces base ten.
  A number that is confidently wrong is worse than no number, which is the whole point of the feature.

## Not done

Option 2 from the issue — budgeting concurrency by expected cost — is **not** implemented and I don't
think it should be. It is a throughput optimisation; the false failures were a *correctness* bug, and
the frame budget removes them whatever the concurrency. Capping long scenarios would also slow the
tier down on an idle machine, which is most of the time.

**210 tests pass, `tools/check.py` clean.** Not reviewed yet — needs `/code-review` before merge.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4251,14 +4251,29 @@ blew 300-600s walls, and the tier meant to be the pre-playtest safety net was th
 All eight passed under `--jobs 1`.
 
 **The work is frame-bound, so the budget is frames.** `deadline` is now read as what it always
-meant -- how much WORK a scenario may do -- expressed as `deadline * fps` frames at its own target
-rate. Under contention a run takes longer in wall time and its verdict does not change. The
-identical arithmetic applies on an idle machine, so nothing about the healthy case moved.
+meant -- how much WORK a scenario may do -- in frames rather than seconds. Under contention a run
+takes longer in wall time and its verdict does not change.
+
+**The rate is FLOORED at the declared target, not fixed to it.** The emulator outruns its own
+target: `titlecard` measures 877fps against a 240 target, so a flat `deadline * target` would have
+been ~3.6x stingier than the wall deadline it replaces -- and `llm`, whose own comment says *"at
+240fps a frame budget would be 4x too impatient"*, would have started failing. The budget therefore
+uses the best rate the run has actually achieved, floored at the declared one: that reproduces the
+old allowance on an idle machine, while contention can only make the budget LARGER. **The bad
+direction is the one that cannot happen.** It is sized on the DECLARED rate too, never on a `PT_FPS`
+override -- watching a run at 60fps must not quarter its work allowance.
 
 **A frame budget alone would hang forever**, because a wedged emulator emits no frames and so never
 reaches any budget. Two kill paths, and both name a real fault rather than a busy laptop:
 **OVERRUN** (more frames than budgeted -- genuinely not finishing) and **STALL** (the frame counter
 stopped -- mGBA is wedged). `PT_MAX_WALL_S` stays as a last-resort net for neither.
+
+**STALL is the one piece that is still a wall clock, and the first cut set it wrong.** Progress is
+read from stamped log lines, but the harness goes deliberately silent inside long waits -- the
+largest is 9000 frames, which at the ~15fps contention floor is **600 SECONDS**. A 180s default
+would have killed healthy contended runs as "wedged", recreating the exact bug this entry is about,
+one layer down. The default is 1800s, and a test pins it against the largest `waitFor` in
+`harness.lua` so the two cannot drift apart.
 
 Every run now prints `throughput: N frames in Ns = Nfps (target Nfps)`. The harness already stamped
 every line with its frame, so measured fps was free the whole time; printing it is what makes a
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4242,6 +4242,36 @@ different guards in `check.py`, both times from a sloppy source-slicing edit tha
 instead of replacing it. `check_no_shadowed_definitions` now rejects it across `tools/`. Same family
 as *"A guard that stops guarding fails silently"*: the failure is indistinguishable from working.
 
+**A deadline that measures the MACHINE is not a deadline (2026-09-02, #345)**
+A `SUITE=all` sweep tabled **eight chapters as FAIL**. Nothing had failed. `SUITE=all` puts nearly
+every long scenario in the repo in one `canonical` block, so four multi-minute mGBA processes ran
+concurrently against `-j8` ROM builds at roughly **15fps against a 240fps target** -- 16x slower
+than solo. Every deadline in `matrix.yaml` was WALL-CLOCK, so scenarios that pass alone in 25-70s
+blew 300-600s walls, and the tier meant to be the pre-playtest safety net was the one that broke.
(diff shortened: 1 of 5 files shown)


NicoviGH:
## All four addressed — and the first one was the whole point

**You caught me moving the bug rather than removing it.** STALL reads progress from stamped log
lines, but the harness goes deliberately silent inside long waits — the largest `waitFor` in
`harness.lua` is **9000 frames**, which at the ~15fps contention floor this PR exists to fix is
**600 seconds**. My 180s default would have killed healthy contended runs as "mGBA is wedged":
the same machine-measuring bug, one layer down, showing up only in the sweep it was meant to save.

| Finding | Resolution |
|---|---|
| STALL is a wall clock over a log with no heartbeat | Default 180s → **1800s**, derived from the largest `waitFor` at the 15fps floor, and a test parses `harness.lua` to pin it so the two cannot drift |
| `budget = deadline*fps` assumes the emulator runs *at* target | Budget now uses the **best rate actually achieved, floored at the declared one** |
| `PT_FPS` scaled the work budget | Budget is sized on the **declared** rate (`MX_FPS`), passed separately from playback fps |
| `PT_STALL_S`/`PT_MAX_WALL_S` unnormalized in the cache key | `RUN_SH_DEFAULTS` normalizes them like `PT_DIFFICULTY`, with a test parsing `run.sh`'s own `${X:-N}` to prove the files agree |

**On the rate:** you were right that 877-vs-240 makes a flat budget ~3.6× stingier than the wall
deadline it replaces, and `llm` was the sharp case — its own comment says *"at 240fps a frame budget
would be 4x too impatient"* and I was about to impose exactly that from outside. Flooring instead of
fixing gives the old allowance on an idle machine, and contention can only make the budget **larger**.
The bad direction is now the one that cannot happen.

**Re-verified in-engine after the arithmetic changed** (not just re-run the tests):

| Path | Result |
|---|---|
| healthy | PASS, exit 0 |
| OVERRUN | budget **1267 frames** — 1s at the rate actually achieved, not a stingy 240 — and it still fires: `OVERRAN its frame budget (3803 > 1267 frames)`, exit 1 |

`matrix.yaml` is unchanged by that test; it is restored in the same command that edits it.

**216 tests pass, `tools/check.py` clean.**

PR #359, opened 2026-09-03
difficulty: AI behaviour is MEASURED and REPORTED, not weighted into threat (#344)

### LINKED ISSUE #344: Factor AI behaviour into the difficulty metric (#335 stretch)
#335's stretch goal, split out so #335 can close on its three delivered scope items.

## The gap

#48's threat and clear-load metrics treat all threat as equally imminent. AI splits it in two, and the vocabulary is already in ch05's YAML:

> *"what she adds is threat the player chooses to walk into rather than threat that comes to them."*

- **Threat that comes to you** — `pursue`, `charge-after-one-turn`, `pillage-then-pursue`
- **Threat you walk into** — `never-move`, `guard-tile`

A never-move unit contributes nothing until the player enters its range, so turn-1 `threat/slot` **overstates a static map and understates a mobile one**. #171 already splits line / reinforcements / convertibles; this is the same idea along the behaviour axis.

## Why it is worth doing

This is the metric that would have caught the drift #335 fixed. ch04 fielded 78% pursuing units against a half-static twin and still measured x1.00, because the numbers only ever saw stats and weapons. The AI is now DERIVED per unit (`donor:` -> the twin's four `.ai` bytes) and its behaviour family is already computed for `make chapter CH=chNN`, so the input this needs exists — it just is not folded into the difficulty math.

## Sketch

`tools/difficulty.py` already resolves every enemy's emitted AI via `enemy_ai_bytes`, and `chapter_status.AI_BEHAVIOUR` maps the AI2 script index to a family. Weight each unit's threat contribution by family when computing the turn-1 figure, and report the two totals separately rather than collapsing them.

Wants a measurement pass against the twins before any weighting is picked — the point is to match vanilla's shape, not to invent a formula.

## Not in scope

The chapter-level `ai_divergence:` allowlist #335 originally proposed is **deliberately not built** — see `docs/decisions.md` -> "A donor carries its AI, so there is nothing left to allowlist". A donor carries its AI, so there is no authored second opinion for an allowlist to referee, and deliberate departures are declared per-unit via `ai_override:` with a mandatory `why`.

NicoviGH:
Closes #344.

The issue asked for a **measurement pass against the twins before any weighting is picked**. The
measurement says: don't build the weighting.

## Vanilla has no single shape to weight against

Red-force behaviour per twin, from each chapter's own `UnitDefinition` `.ai`:

| twin | n | mobile | static |
|---|---|---|---|
| FE8 Prologue | 3 | 33% | 33% (+1 unclassified) |
| FE8 Ch1 | 10 | 70% | 30% |
| FE8 Ch2 | 9 | **100%** | 0% |
| FE8 Ch3 | 10 | 0% | **100%** |
| FE8 Ch4 | 23 | 56% | 43% |
| FE8 Ch5 | 23 | 52% | 47% |
| FE8 Ch6 | 27 | 33% | 66% |
| FE8 Ch13 | 52 | 61% | 38% |

Behavioural parity can only ever mean *this chapter against its own twin*.

## And ours already equals its twin, everywhere

**0 points of difference on all six active chapters**, every enemy resolving, 0 failures. Not luck:
**#335 made AI derived** — `enemy_ai_bytes` returns the donor's bytes — so our shape *is* the twin's
by construction. A weighting would apply identical weights to both sides of a ratio that is equal by
definition and report x1.00 whatever happened. **It cannot catch anything.** The ch04 "78% pursuers
against a half-static twin" case is history: deriving AI fixed it, measuring it did not.

The only remaining divergence channel is `ai_override` — three in the campaign, each with a mandatory
`why`, none a silent shift.

## What this ships instead

- **The split in the open.** `make chapter` prints `shape N come to you / N you walk into`, beside
  the roster it describes. ch05: `12 come to you / 11 you walk into`. A future divergence is visible
  rather than latent, and the metric stays about stats — which is what it can actually measure.
- **`AI_BEHAVIOUR` completed from the decomp.** It named 8 of the 19 AI2 scripts
  (`gAi2ScriptTable`, `cp_data.c:1548`, read by `cp_script.c:90` as `gpAi2Table[0][ai2]`), so eleven
  reported as bare numbers — including `0x0A`, which **vanilla's own Prologue fields**. ch00 now
  reports `1 / 1 / 1 unclassified`.
- **Unnamed stays unclassified.** The eleven entries named after bare addresses are *not* given
  invented names or bucketed. Calling an unknown script "static" because it is unnamed would be a
  guess wearing a measurement's clothes, and it would land on the side that lowers the number.

## The test pins the map against the decomp

Against `cp_data.**c**` — the tracked source — never the generated `.s`, which is a build artifact and
is **not at HEAD at all** (the first cut read it and exited 128, the same class of mistake as reading
the built decomp tree). It parses the **designated initialisers**, so a table with a hole cannot
silently shift the indices.

**41 tests pass in `test_chapter_status.py`, `tools/check.py` clean.**

Not reviewed yet — needs `/code-review` before merge.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -4270,13 +4270,21 @@ latent -- and the metric stays about stats, which is what it can actually measur
 
 **A corollary, and the one real gap the measurement found.** `AI_BEHAVIOUR` named 8 of the decomp's
 19 AI2 scripts (`gAi2ScriptTable`, `cp_data.c:1548`, read via `cp_script.c:90` as
-`gpAi2Table[0][ai2]`), so eleven reported as bare numbers -- including `0x0A`, which **vanilla's own
-Prologue fields**. The map is now complete for every index whose decomp SYMBOL states its behaviour,
-and the eleven named after bare addresses are left **unnamed and UNCLASSIFIED** rather than guessed
-at: bucketing an unknown script as static because it is unnamed would be a guess wearing a
-measurement's clothes, and it would land on the side that lowers the number. A test pins the map
-against `cp_data.c` -- the tracked `.c`, never the generated `.s`, which is a build artifact and not
-at HEAD at all.
+`gpAi2Table[0][ai2]`), so eleven indices reported as bare numbers -- including `0x0A`, which
+**vanilla's own Prologue fields**. Thirteen are named now: every index whose decomp SYMBOL states
+its behaviour. The **six** the decomp itself gives only an address (0x08, 0x09, 0x0A, 0x0B, 0x0D,
+0x0F) stay **unnamed and UNCLASSIFIED** rather than guessed at -- bucketing an unknown script as
+static because it is unnamed would be a guess wearing a measurement's clothes, and it would land on
+the side that lowers the number. A test pins the map against `cp_data.c` -- the tracked `.c`, never
+the generated `.s`, which is a build artifact and not at HEAD at all.
+
+**"Known but neither" is not the same fact as "unknown", and the first cut collapsed them.** Three
+NAMED families -- `pillage-escape`, `escape`, `attack-walls-snags` -- are neither threat that comes
+to you nor threat you walk into: the unit moves, but not at the player. Leaving them to fall through
+to `unclassified` said *we do not know* about scripts the decomp names outright, and `ai2 = 0x05`
+alone appears 120 times in the reference files the parity twins already read. They get their own
+bucket, **on its own errand**, and a test asserts that no named family can ever fall through to
+unclassified again.
 
 **A deadline that measures the MACHINE is not a deadline (2026-09-02, #345)**
 A `SUITE=all` sweep tabled **eight chapters as FAIL**. Nothing had failed. `SUITE=all` puts nearly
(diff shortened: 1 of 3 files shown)


NicoviGH:
## All three addressed

**The first one had real teeth and you were right about why.** `pillage-escape` (0x05), `escape`
(0x0C) and `attack-walls-snags` (0x0E) are *named* families that were falling through to
`unclassified` — which says "we do not know" about scripts the decomp names outright. With `ai2=0x05`
appearing 120× in the reference files the twins already read, a looting thief would have printed as
unknown.

They now have their own bucket — **`on its own errand`**: the unit moves, but not at the player. And
a test asserts that **no named family can ever fall through to unclassified again**, which is the
invariant I stated in the comment and then failed to hold.

| Finding | Resolution |
|---|---|
| Named families in neither bucket | New `AI_OWN_ERRAND` bucket + an invariant test over every entry in `AI_BEHAVIOUR` |
| "eleven bare-address entries" is wrong — there are six | Fixed in all six places. Eleven was the count of previously-*unmapped* indices, a different quantity, and I conflated the two |
| Stale `cp_data.s:3245` citation | → `cp_data.c:1548`. Embarrassing: that is the untracked build artifact this very change warns against reading |
| `SYMBOLS` omits 0x01, 0x02, 0x07 | All thirteen names pinned now — a decomp rename there would have mis-bucketed units past a green suite |

Rendered after the change: ch05 `12 come to you / 11 you walk into`, ch00 `1 / 1 / 1 unclassified`
— still the live `0x0A`, correctly still unknown.

`tools/check.py` clean, suite green.

PR #364, opened 2026-09-04
ch06: host The Maer Monster on slot 7 (#26)

NicoviGH:
Hosts ch06 "The Maer Monster" on slot 7. The two commits already on this branch declared the host slot; this PR adds `inject_ch06`, the injector that fills it.

```
ch06 map (obj1=255 pal=118 cfg=119 layout=120) hosted on chapter 7;
defeat_boss goal + DefeatBoss(Nerra) WIN wired, PREP deploy cap 10
+ 24 line + t4:3 hard-only + 2 boat(s)
```

Parity: **x1.00 threat / x1.00 clear-load** against FE8 Ch6, role check clean, battlefield dynamics 1:1 — so the chapter flips to `status: active`.

## The asset table hit its 8-bit ceiling

`gChapterDataAssetTable` is addressed by **nine u8 fields** of `ROMChapterData` — the eight `map.*Id`s plus `mapEventDataId` — so **255 is the last index the engine can reach**. Vanilla ships 236 entries and ch00–ch05 had appended 19: **255 exactly.** ch06's four more ran the table to 260.

The only thing that caught it was agbcc's `-Werror` on an implicit truncation in generated code. Nothing of ours would have: every id is derived per build and written into `chapter_settings.json` by the injector that registered it, so the numbers stay self-consistent right up to the truncation. The ROM would have built clean and ch06 would have drawn whatever lived at `258 & 0xFF`.

**There is no free list to find.** Reading only the `map.*` fields reports 60 unreferenced entries — and every one of them is a `ChapterEventGroup` reached through the ninth field. Counting all nine, vanilla's 236 entries are referenced without exception.

So a campaign asset now **claims a slot belonging to a vanilla chapter our ROM can never load** — the same argument that already lets us strip host-slot event lists, squat dead message blocks, repurpose unreferenced scripts and take unnamed `gCharacterData` gaps. Appending still happens first, so **every id ch00–ch05 built with is byte-identical**; only the assets that could not fit at all moved. It reserves everything vanilla slots 0–25 touch (we host on 1–9) and leaves 118 reclaimable entries.

What it picked is the entry's own footnote — ch06's palette, tile config and layout took vanilla's `Ch13EphraimMap` / `Ch13EphraimMapChanges` / `Ch13EphraimEventData` slots. ch06's geometry is a retile of Ch13 Ephraim and its deploy tiles come from that group's own unit table, so it now occupies its donor's runtime slots as well as its shape.

Reasoning: `docs/decisions.md` → *"The asset table is addressed by a u8, so ch06 RECLAIMS a slot rather than appending"*.

## Two silent inheritances, both now written

- **Slot 7 ships `initialFogLevel: 3`** — vanilla Ch7 is fogged. ch06 declares `fog: none` and its whole route-puzzle design depends on it. Second ADR, because unlike the goal ids, the battle grounds, the difficulty numbers and `.traps`, fog still has **no pass** that writes every hosted chapter's declaration. The general pass is owed.
- **Vanilla Ch7 carries two ballistae.** `traps: []` is declared in the YAML — a frozen lake has no ballista emplacement.

## The boats

Green `CLASS_FLEET` units, not painted scenery, so they can be killed — that clock is the chapter's difficulty. Their AI is derived like every enemy's: each boat's `donor:` is vanilla Ch6's boss tile, so both inherit `GuardTileAI`. **Load-bearing, not decorative** — `CLASS_FLEET` has `baseMov 3`, so a hull on the zero-filled default AI (ActionInRange + MoveToEnemy) would sail off its own outcrop at the party, taking the pocket, the door and the entire fuse with it.

## Two bugs ch06 surfaced rather than caused

- `inject_names` ordered appended name plates by **pid**, which only worked while the highest pid happened to own the highest message id. `gMsgTable` is dense, so the boats' 0xD4D/0xD4E on pids 0xb4/0xb5 tried to append ahead of the moose's 0xD4C.
- `_asm_table_word_index` matched a raw line, so it would find every appended asset and miss exactly the reclaimed ones.

## Deferred on purpose

All declared with empty text: the boarding Talks and their rewards, Messie's boss-death cutscene, the opening and ending scenes, and the merfolk reskins (`skin:` is intent; the wiring lands in `campaign.yaml`). The ending parks on the dev placeholder until ch07 hosts, exactly as ch05's did.

⚠️ **This does not close #360's second open question.** One instrumented run on `ch06boot` can now settle that enemy AI paths to a single melee tile and queues rather than stalling — but "Talk fires from the door" needs a Talk, and the boarding pass is where that lands.

## Verification

- `python3 tools/check.py` — drift check clean (lint + every unit suite)
- `python3 tools/verify_text.py` — 3404 messages, 0 runaway
- `make CAMPAIGN=rime-of-the-frostmaiden fireemblem8.gba` — builds
- `make difficulty CH=ch06` — PARITY (within band), role check clean
- New `AssetTableCeiling` suite: 8 tests pinning the ceiling, the pool, the reserve, the append/reclaim split, and the loud failure when it runs out

Not yet run in-engine — the instrumented playtest is a deliberate act and is Nicolas's call.


NicoviGH:
## `/code-review` found two real bugs. Both fixed in 507e289.

**1. The boat pids collided with ch04's wolf pack.** The boats took `0xb4`/`0xb5`, and `CH04_PACK_PIDS` already owns `0xb0/b1/b2/b4/b5`. A name plate is written into `gCharacterData[pid - 1]` — **one global row per pid, no chapter dimension** — so two of ch04's five Mauthe Doogs would have read **"Fishing Boat"** on the unit window.

My constant's comment claimed only `0xB6..0xB9` were taken and didn't know about ch04's five. `0xB0..0xBA` is in fact full. Moved to `0xbb`/`0xbc`, the next unnamed `0x255` gaps.

The prose was the bug, so prose isn't the fix: `assert_named_raw_pids_are_exclusive` now **discovers** every raw pid the module claims — string, tuple or dict constants alike — and fails the build when a *named* pid is claimed by two chapters. It deliberately permits a shared **generic** pid: ch05 and ch06 both spend `0x80` on trash, and ch03's grell shares `0xb7` with ch04's mogall quite legally, because `gDefeatTalkList` is keyed by chapter and `gCharacterData` is not. The guard reproduces the original collision as a test.

**2. `item_drop` was appended even when already in `inventory`.** ch06 is the first chapter whose YAML names the drop in both places, so three enemies emitted a spare copy — the halberd Fighter shipped `{AXE_IRON, AXE_HALBERD, AXE_HALBERD}` against a donor carrying two items. `make difficulty` read **PARITY** throughout, because the parity model prices the YAML rather than the rows the injector emits.

`_items_with_drop_last` **re-orders** rather than merely de-duplicating — "last" is the part the engine reads, so an inventory listing the drop first would otherwise drop the wrong item. Fixed in `ch05_enemy_rows` too, where the same shape sat untriggered because ch05 declares no drops at all.

**Verified against the emitted decomp rather than the YAML** — which is the whole lesson of finding 2:

```
.items = { ITEM_AXE_IRON, ITEM_AXE_HALBERD },   shark-rider-halberd  (donor 13,6 carries 2)
.items = { ITEM_BLADE_IRON },                   merfolk-blade-drop   (donor 15,10 carries 1)
.items = { ITEM_STAFF_MEND, ITEM_ELIXIR },      lamia-mender         (donor 10,0 carries 2)

[0xb4 - 1] .nameTextId = 0x255   [0xbb - 1] .nameTextId = 0xD4D
[0xb5 - 1] .nameTextId = 0x255   [0xbc - 1] .nameTextId = 0xD4E
```

Also took the review's doc nit: the ADR claimed *"every id ch00–ch05 built with is byte-identical"*, which isn't quite true — `inject_prologue` runs **after** `inject_ch06` in `main()`, so the prologue's layout moves to a reclaimed slot. Harmless (ids are derived per build) but a false invariant to state, so it's reworded.

`make check` clean · `verify_text` 3404 messages, 0 runaway · ROM builds · 624 tests pass.

NicoviGH:
## In-engine verification: gate green, ch06 loads clean — plus a harness bug this found

### Pre-merge gate: 15/15 PASS, 6m14s, **0 cached**

Every scenario genuinely re-ran, so this is a real result rather than a cache artifact. That matters here specifically: the asset-table reclaim rewired how *every* shipped chapter resolves its map, palette, tile config and layout, and the prologue's own layout moved into a reclaimed slot — which every `canonical` scenario boots through. Nothing moved.

### ch06 load test: PASS, and the roster is exactly as authored

```
blue: 10   green: 2   red: 24
blue tiles (7,1) (6,0) (5,1) (9,1) (4,0) (8,0) (6,2) (4,2) (6,1) (10,0)
```

That is `deployment.deploy_slots` in order — the chapter fields its full cap of 10. And:

| | |
|---|---|
| Boats | `0xbb` (17,12) · `0xbc` (4,17), **19 HP** each = `CLASS_FLEET` baseHP, the hull the sinking clock is tuned against |
| Boss | `0x4b` = `CHARACTER_NOVALA` at (10,12), **28 HP** — Nerra inherits Novala's real line, which is the point of `ENEMY_BASE_SLOT` |
| Line | 24 reds |
| Hard wave | **absent**, correctly — turn-4 Difficult-only, and this ran on Normal. `EventScr_LoadReinforceHardMode` doing its job |

### The harness bug (baaad85)

The first ch06 load test reported **2 blue of 10**, and I initially called it a ch06 defect. It was not — the diagnostic was lying.

`INSPECT.units` walks `gUnitArrayBlue` directly, and `bootToMap` returns on `player_map_idle`, which can be true while the chapter's beginning script is still running. On a chapter whose opening is a bare `LOMA` / `LOAD1` / `CALL(prep)` spine there is almost nothing between the Fight press and that state, so the dump read a party PREP had not finished redeploying. The screenshots 120 frames later showed the lord standing on the map at full HP, which is what disproved it.

**Every chapter before ch06 hid this, in the way that makes a latent bug expensive.** ch05's opening plays four backdrop scenes, a join and a monologue *after* prep, so its dump lands ~15,000 frames in, long settled. So the bug was invisible on chapters that have scenes and fires on chapters that do not — which is precisely what this scenario exists to test. **ch07 and ch08 would have inherited it.**

It matters beyond a flaky number because `docs/adding-a-chapter.md` tells people to trust this dump *over* a screenshot (FE8 draws a map sprite offset upward, so a unit reads a row high). That advice is right — the dump just has to be true for it to be safe. Fixed by waiting on the same predicate `mapfull` already waited on; no new top-level Lua local, so `harness.lua` stays clear of the 200-local ceiling (2 free).

ADR: `decisions.md` → *"A load-test that reads the roster before PREP settles is a diagnostic that LIES"*.

### Still not proven

Enemy AI queueing at the boat door. That needs a scenario watching several enemy phases, and it lands with ch06's owed `playtest:` block — it is not something `mapshot` can answer.

Issue #367, opened 2026-09-04
Does the difficulty model have a spatial element we designed and never wired?


NicoviGH:
Split out of #26 (2026-09-04). Not to be worked inside the ch06 fix — it needs a fresh look.

## The question

`docs/decisions.md` → *"Difficulty is checked in fidelity tiers"* says `make difficulty` is **aspatial by construction** — it sums per-unit offense/durability as if every fight were 1-on-1 on open ground — and is therefore *"blind to terrain, unit placement, AI behavior, fog, and the weapon triangle."*

But we have since built spatial machinery, and it is not clear whether some of it was always meant to feed the difficulty math rather than sit beside it:

- `map_placement_preview.foot_reach` — now **occupancy-aware**, with per-class cost tables and contested/uncontested arrival (`reached_on_contested:`, landed 2026-09-04 on #26).
- `terrain_divergence` and the pocket/door modelling.
- `chapter_status.ai_shape` — the three-way pursuer/striker/statue split, which is behavioural rather than spatial but was the missing half of the same picture.

**Is there a spatial element of the difficulty model that was designed and never wired, or that exists and we have stopped reading?** Nicolas' recollection (2026-09-04) is that one was designed. Check the issue history for #45 and #48 before assuming not.

## Two findings from #26 that belong in the same look

1. **The player side is modelled at base level, the enemy side is not.** `difficulty.player_combatant` resolves the cast *"at base level"*; `difficulty.enemy_combatants` autolevels each entry to its declared level. So every chapter's parity is levelled-enemies vs a level-1 party. The ratio still cancels to FIRST ORDER — the same understated party appears top and bottom — but `kills_per_round` and `durability` are threshold functions (doubling breakpoints, integer rounds-to-kill), so a party far below the enemies' level can saturate against both sides and flatten the ratio toward x1.00 regardless of what is underneath. **The error is smallest at ch00 and grows with every chapter.**

2. **Nothing anywhere states an expected party-level band per chapter.** It is not needed for parity (see above) but it is the floor every ABSOLUTE claim needs — "can this unit survive that trip", "is this fuse long enough for a real party". Its absence is how a ch06-era flier got assessed off a level-1 stat line during #26. A documented planning band in `docs/fe8-pacing-reference.md` is the cheap version; a growth model is the expensive one, and probably not warranted.

## Not in scope

The ch06 fix on #26. That stands on the parity-ratio ground, which is sound, and on tier-3 verification by play.

[...]

PR #368, opened 2026-09-05
difficulty: print mirror% beside every parity ratio (#367)

NicoviGH:
Proposal 1 off #367, and Nicolas's first Next task.

## The problem

The parity ratio is an aggregate over stats, so it cannot distinguish a chapter that transcribes its twin unit for unit from one that composes a different force arriving at the same per-slot pressure. Both read x1.00.

ch06 is the live case: a perfect **x1.00 threat / x1.00 clear-load** against vanilla Ch6 — on the one chapter whose sinking clock measured wrong in *both* directions. Both facts are true and not in tension, because ch06 reproduces **100% of Ch6's 27-unit force**: its score is a checksum on the donor pipeline, not evidence about the chapter.

## What this adds

`difficulty.mirror_share(chap)` — the share of the **twin's** force reproduced exactly, printed beside every ratio:

```
  chapter                reference     threat/slot     clear-load/slot   mirror  verdict
  CH0 ch00-prologue-a-da FE8 Prologue   0.9 (x1.13)      0.3 (x1.00)       33%     OK  [locked]
  CH1 ch01-the-iron-trai FE8 Ch1        9.3 (x0.89)      3.5 (x0.97)       30%     OK  [locked]
  CH2 ch02-cold-welcome  FE8 Ch2        6.5 (x1.00)      3.8 (x1.00)       100%    OK  [locked]
  CH3 ch03-the-termalain FE8 Ch3        3.5 (x1.03)      2.2 (x1.00)       90%     OK
  CH4 ch04-the-white-moo FE8 Ch4        9.2 (x1.15)      3.1 (x1.19)       61%     OK
  CH5 ch05-the-elven-tom FE8 Ch5       14.5 (x1.04)      7.5 (x1.03)       52%     OK
  CH6 ch06-the-maer-mons FE8 Ch6       14.9 (x1.00)      7.5 (x1.00)       100%    OK
```

and in the per-chapter report, which says out loud what a high share means:

```
  mirror: 100% of FE8 Ch6's force reproduced exactly (27 of 27 bodies, class + level)
          -- at this share the verdict is largely a CHECKSUM on the donor pipeline,
             not evidence about the chapter.
```

**The trend is the finding, not the ch06 cell**: mirror% has climbed as the donor pipeline improved, so every gain in derivation fidelity made the gate more tautological — and there was no number watching that happen.

## Definition (each choice argued in the ADR)

- **(class, level)**, not class — an L9 fighter is not the twin's L1 one.
- **Multiset intersection** — a doubled body cannot score twice against a twin that fields one.
- **Denominator is the TWIN's body count** — fielding more than the twin never reads above 100%.
- **Both sides count line AND reinforcements** — the twin's curated arrays already include its waves.
- **Nothing dropped for carrying no modeled weapon** — this measures the force's SHAPE, so a staff-only healer is still a body. (Hence ch06 printing 25 modeled enemies and 27 mirrored bodies on adjacent lines, on purpose.)

**It does not gate.** A high mirror is not a defect and a low one is not either; it says how much a verdict is worth.

## Verification

- ~45 new tests, each written first and watched fail.
- `tools/test_difficulty.py` **248 OK** · `tools/test_build_campaign.py` **627 OK** · `tools/test_check_chapter_schema.py` **28 OK** · `make check` green, drift clean.
- **Output diff** over all **nine** chapters (`--chapter` and `--lord-floor`) and **all four** curve reports (authored + tutorial/normal/difficult), before vs after: **only ch02's numbers move.**

The mirror numbers reproduce #367's independently-computed spike (33/30/**78**/90/61/52/100) — except ch02, and that disagreement is the finding below: #367's spike counted our opening board against the twin's full arrays, exactly as the shipped metric did.

No ROM build, no emulator, no playtest run.

---

## What this turned up: ch02 was never counting its own wave

Defining mirror% forced the question the ratio had never been asked — *what is a body?* — and the answer was that our side and the twin's were counting different forces. Full write-up in the [review-response comment](https://github.com/NicoviGH/manchego-stars/pull/368#issuecomment-5548866510); the short version:

`chapter_units` read only `enemy_units`; `vanilla_enemies` reads every red `UnitDefinition` in the twin's curated arrays, **reinforcement waves included**. ch02 declares its rear-raider pair under `reinforcements:` (with `levels: [2, 3]` and no `level:`), so its verdict compared **7 of our bodies against 9 of vanilla's**, both of them modeled at L1.

| ch02 | shipped | corrected |
|---|---|---|
| threat/slot | 5.3 (**x0.81**) | 6.5 (**x1.00**) |
| clear-load/slot | 3.0 (**x0.79**) | 3.8 (**x1.00**) |
| mirror | 78% | **100%** |
| bodies ours vs twin | 7 vs 9 | 9 vs 9 |

Verdict was `OK` before and after — it never flipped a gate, which is how it survived four chapters on a `balance_locked` chapter. **So there are two copy-x1.00 chapters, not one**, and ch06 is not the first: ch02 was already a transcription in 2026-08 and the instrument was hiding it.

Fixed at the root. Three definitions now live on `build_campaign`'s desk, which owns the chapter schema, because *four separate gates* key off the answer and each had its own copy of the question:

- **`ENEMY_ROSTER_KEYS`** / **`chapter_roster_entries`** — what units does this chapter field.
- **`entry_body_levels`** — how many bodies does one entry place, and at what levels.
- **`_entry_combatants`** (in `difficulty`) — the one entry expander, replacing three near-copies. That duplication is exactly how `levels:` came to be honored in one reader and defaulted to L1 in the others.

Nine further readers and guards were brought onto them. Four were real holes, all on `balance_locked` chapters and all silent:

| guard | what got through |
|---|---|
| `unmodeled_enemies` | a staff-only boss under `reinforcements:` raised no BOSS DROPPED, so `--check` stayed green |
| `role_findings` | a boss there contributed threat but could never be graded per-unit |
| `_personal_line_route_violations` | a boss there could declare `personal:` with no route into the ROM — it MEASURES fixed and PLAYS naked, #284's exact failure |
| `unregistered_raw_pid_bosses` | a raw-pid boss there got no `RAW_PID_LEVEL_SOURCES` row while `_our_base_level` modelled it malus-immune |

## Review

Five `/code-review high` rounds; the write-ups are in the thread. Rounds 1–3 each found a live defect (including one this branch introduced and I had missed by diffing only ch00–ch06 — ch07/ch08 are `status: planned`, but the tool still grades them). Rounds 4 and 5 found no live break, only latent gaps one ring further out each time, which is where I stopped.

Part of #367.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -8525,13 +8525,16 @@ past. That is the `drop_staff` flag, and the output diff is what caught it: ch06
 moved 7.54 → 7.41 and three lords' worst matchups changed, because the menders had silently left
 the field.
 
-**A per-unit question is answered per unit TYPE, not per body.** The shared expander yields one
-body per copy, and `role_findings` took it as-is — so ch08's `count: 4` ice-troll read as *"4 units
-flagged is_boss (ice-troll, ice-troll, ice-troll, ice-troll)"* and repeated every warning four
-times. Whether a unit's threat is an outlier does not depend on how many of it stand on the map,
-and `role` is a `--check` gate arm, so that would have redded CI on the first locked chapter with
-a multi-copy boss entry. Hence `distinct`: bodies of one entry collapse unless `levels:` actually
-makes them different units.
+**A per-unit question is answered per unit TYPE, not per body.** This one was self-inflicted and
+caught before it landed: consolidating on the shared expander made `role_findings` yield a row per
+BODY, and ch08's `count: 4` ice-troll immediately read as *"4 units flagged is_boss (ice-troll,
+ice-troll, ice-troll, ice-troll)"* with every warning repeated four times. (The old code was
+correct here — it expanded per distinct TYPE — so this is a regression the consolidation
+introduced, not a bug it found.) Whether a unit's threat is an outlier does not depend on how many
+of it stand on the map, and `role` is a `--check` gate arm, so it would have redded CI on the first
+locked chapter with a multi-copy boss entry. Hence `distinct`, which collapses bodies of one entry
+unless `levels:` makes them genuinely different units — and a boss CENSUS that counts entries, since
+`distinct` still splits one entry by level and every row carries that entry's id.
 
 **And the output diff has to cover every chapter the tool reads, not the ones the change was
 about.** The first diff ran ch00–ch06 and reported "only ch02 moves" — which was true of those
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -8428,9 +8428,9 @@ the same number, and nothing in the report distinguished them.
 `mirror_share` prints, beside every ratio, the share of the **twin's** force the chapter
 reproduces exactly. Measured across the campaign:
 
-| ch00 | ch01 | ch02 | ch03 | ch04 | ch05 | **ch06** |
+| ch00 | ch01 | **ch02** | ch03 | ch04 | ch05 | **ch06** |
 |---|---|---|---|---|---|---|
-| 33% | 30% | 78% | 90% | 61% | 52% | **100%** |
+| 33% | 30% | **100%** | 90% | 61% | 52% | **100%** |
 
 **The trend is the finding, not the ch06 cell.** mirror% has climbed as the donor pipeline
 improved — every gain in derivation fidelity made the gate more tautological, and nobody was
@@ -8444,11 +8444,8 @@ What a body IS, and why each choice:
   one.
 - **The denominator is the TWIN's body count**, so fielding more than the twin never reads above
   100%. The question is how much of the twin we reproduced, not how much of ours is borrowed.
-- **Both sides count line AND reinforcements.** The twin's curated arrays already include its
-  waves, so counting only our opening board would compare nine vanilla units against seven of
-  ours and call the gap divergence. (Live case: ch02's two L1 brigands are reinforcements, and
-  its twin fields no L1 brigand — so the honest read is still 78%, arrived at with both waves on
-  the board.)
+- **Both sides count line AND reinforcements**, at each body's DECLARED level. This is the
+  choice that found a bug, and the bug is below.
 - **Nothing is dropped for carrying no modeled weapon.** The pressure metric drops a staff-only
   healer because it contributes no damage; this measures the force's SHAPE, and a healer the twin
   fields is a unit we did or did not reproduce. So the two numbers on the same line count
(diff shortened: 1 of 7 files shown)


NicoviGH:
## Review findings addressed — and the second commit is the interesting one

All four findings from `/code-review high` were verified independently and fixed in `06add25`.

**Finding 1 (`levels:` ignored) was correct, and verifying it found the defect underneath it.**

`_force_signature` ignoring `levels:` was my bug. But `chapter_units` — *the* our-side force builder, which produces the parity verdict itself — reads only `enemy_units`, while `vanilla_enemies` reads every red `UnitDefinition` in the twin's curated arrays, **reinforcement waves included**. ch02 declares its rear-raider pair under the `reinforcements:` key, so its verdict has been comparing **seven of our bodies against nine of vanilla's** and scoring the missing two as divergence. The same entry carries `levels: [2, 3]` with no `level:`, so both bodies were additionally modeled as L1.

ch06 had already hit this and worked around it privately — its Hard-only turn-4 wave is declared inside `enemy_units` rather than under `reinforcements:`, and its own YAML says why:

> "#48's static bar folds that array in on the vanilla side unconditionally, so declaring ours is what makes the two sides count the same force."

That is the general rule, and it was being enforced one chapter at a time by hand.

**Corrected, ch02 is not the campaign's easiest chapter. It is an exact transcription of vanilla Ch2:**

| | shipped | corrected |
|---|---|---|
| threat/slot | 5.3 (**x0.81**) | 6.5 (**x1.00**) |
| clear-load/slot | 3.0 (**x0.79**) | 3.8 (**x1.00**) |
| mirror | 78% | **100%** |
| bodies ours vs twin | 7 vs 9 | 9 vs 9 |

The verdict was `OK` before and after — it never flipped a gate, which is exactly how it survived four chapters on a `balance_locked` chapter. `--mode difficult` moves identically.

**So there are two copy-x1.00 chapters, not one**, and #367's "mirror% climbs as the donor pipeline improves" reads differently: ch02 was already a transcription back in 2026-08, and the instrument was hiding it. mirror% found this on its first day, which is the argument for it.

## Findings 2–4

- **Finding 2 (ADR baked in the wrong ch02 number):** correct, and the rationalization was mine. The ADR now carries the corrected trend table and a section on what the "what is a body" question turned up.
- **Finding 3 (`enemy_reinforcements` unread):** fixed at the root. `chapter_roster_entries` answers *"what is this chapter's force"* once, over `AI_ROSTER_KEYS` — so the parity metric and the AI guard can no longer disagree about which units a chapter fields.
- **Finding 4 (mirror under `--mode`):** mirror% is **mode-invariant by construction** — `mode_stats` re-projects the *same level* through the chapter's malus/bonus, so a mode shifts stats and never a unit's class or level, and the Hard-only waves are folded into both sides unconditionally. That is a property to state, not a parameter to add: `mirror_share`'s docstring says it, the `--mode` banner says it on the row, and a test asserts every mirror reads identically across all three modes.

## Verification

- **Output diff** (`tools/callsites.py` first, then every per-chapter report + both curve reports rendered before and after): **only ch02's numbers move.** Every other row byte-identical, in both modes.
- 9 further tests written first and watched fail; `tools/test_difficulty.py` 224 tests OK; `make check` green, drift clean.

PR #369, opened 2026-09-06
rescue_forecast: which pursuer can actually engage its hull, and when (#367, #26)

NicoviGH:
#367 asked whether the difficulty model has a spatial element that was designed and never wired. A full per-tile danger grid was explicitly cut from scope (Nicolas, 2026-09-05) in favor of the narrower, reusable question every chapter with a rescue clock actually asks:

**for a protected tile (a rescue boat, a defend objective, a fragile NPC), which enemies can get a firing position on it, on what turn, for how much damage per phase, and therefore when does it die?**

`tools/rescue_forecast.py` answers exactly that, reusing three things that already existed on three different desks rather than building a new model: `map_placement_preview.foot_reach` (the contested Dijkstra `reached_on_contested:` already uses), `difficulty`'s stat resolution, and `fe_combat.damage_per_round` (the decomp's own combat math).

## The five pieces

1. `firing_cells(terrain, target, weapon_range)` — every cell within range a foot unit can stand on. FE8 has no line of sight, so this is pure Manhattan distance, not a walk.
2. `arrival_to_cells` — the turn an enemy first stands on any firing cell, walking the **contested** map (turn-1 bodies block). `None` — never engages — is a first-class result, not an error.
3. Damage/phase via `fe_combat.damage_per_round`, target resolved on its real terrain.
4. `sink_band` — HP / damage-per-phase is a **mean**, not a fact (`decisions.md` → *"A rescue clock is a HIT RATE"*), so the sink turn is reported as a band (Wald/inverse-Gaussian first-passage approximation), stated in the docstring as a heuristic width, not a claimed percentile.
5. `concurrent_attacker_cap` — a bipartite matching between attackers and the firing cells each one's range reaches, not a naive sum. ch06's two hulls each have exactly one melee door so this isn't load-bearing there yet, but it's built in for wider targets later chapters will ship.

## Validated against ch06's own measured numbers, not synthetic stand-ins

```
merfolk-thrower  -> boat-east   never reaches a firing cell (weapon range 2)
merfolk-thrower  -> boat-west   never reaches a firing cell (weapon range 2)
ice-crab         -> boat-east   never reaches a firing cell (weapon range 1)
ice-crab         -> boat-west   arrives turn 2   2.82 dmg/phase   sinks turn 6-12 (expect 9)
```

boat-east's javelin-range firing cells compute to exactly `(15,12) (17,10) (17,13) (17,14)` (one melee cell `(17,13)`); boat-west's to `(2,17) (4,18) (4,19)` (one melee cell `(4,18)`) — both match their declared `door`.

**The finding this reproduces, in one number: ch06's declared "sinks the hull on turn 7" (east) describes a unit that never arrives.** `merfolk-thrower` computes to 2.76 dmg/phase against the hull's real fighting stats (19 HP / 6 Def in forest cover / +20 avoid) — respectable — but **cannot reach any east firing cell on the contested snapshot: its own allies cork all four.** That prose lived in `difficulty_note:`, not a schema field, so nothing was asserting it and nothing broke when it stopped being true. Same shape as the already-confirmed #26 bug: the tooling that measures a chapter's clock didn't exist, so the design note was never checked against the map it now sits on.

`ice-crab`'s west forecast (arrives turn 2, sinks turn 6–12, expected 9) matches the instrumented run on #26 — arriving enemy phase 2, attacking from `(4,18)` and only that cell for five phases — and lands inside a declared fuse of 8.

## The guard is advisory, not a gate — on purpose

`check.check_rescue_fuse_forecast` prints (never appends to `fail`) when a declared `rescue_pursuers:` id can't reach a firing cell for any of its chapter's `rescue_boats`. **This PR does not touch ch06's YAML to make the guard pass** — the finding is real and the fix (move the pursuer, or re-declare which hull is the east clock) is Nicolas's design call, not a bug in the tool. `chapter_status.loose_ends` surfaces the identical finding for `make chapter CH=ch06`, sharing `check._fuse_forecast_findings` so the two never drift into two different opinions about the same chapter. Flipping reachability to a hard gate is the natural follow-up once ch06's east pursuer is resettled.

## Same code path, two latent bugs fixed alongside it (`map_placement_preview.py`)

- `units_reaching` iterated `chapter['enemy_units']` only, though its docstring claimed reinforcements were included — true only because ch06 happens to keep its Difficult-only wave inside `enemy_units`. Now reads `build_campaign.chapter_roster_entries`, every roster key.
- `enemy_bodies` never read past `enemy_units:` at all, so a `reinforcements:`/`enemy_reinforcements:` entry's body was invisible to the contested Dijkstra's blockers regardless of timing — and simply widening the loop without also fixing the timing test would have flipped the bug rather than closed it, since those keys carry `trigger_turn`, not `arrives_turn` (testing `arrives_turn` alone reads an absent field as falsy, i.e. turn-1 — exactly backwards). The one correct predicate was already living correctly, but only, inside `difficulty.chapter_enemy_groups`; it's now `build_campaign.entry_is_turn1(key, enemy_def)`, and both readers share it.

## Verification

- Output diff over **all nine chapters** (ch00–ch08 — a previous PR in this area got burned diffing only seven): `enemy_bodies`, `units_reaching` (against each chapter's own `rescue_boats`, `[]` where none declared) and `difficulty.chapter_enemy_groups` — **byte-identical before and after, on every chapter.** ch02 is the only chapter using a `reinforcements:` key today, and its wave was invisible to `enemy_bodies` both before (never read) and after (correctly excluded as not-turn-1); no chapter besides ch06 declares `rescue_boats`. `chapter_status.report()` moves on exactly one chapter (ch06), by exactly one line — the new advisory loose-end.
- `tools/test_rescue_forecast.py` (25), `tools/test_check_rescue_fuse_forecast.py` (12), `tools/test_map_placement_preview.py` (28, +new), `tools/test_build_campaign.py`, `tools/test_chapter_status.py` — all green, each test written first and watched fail.
- `make check` — green, drift clean, advisory finding prints as designed (not gating).
- No ROM build, no emulator, no playtest run.

ADR: `docs/decisions.md` → *"A rescue-fuse FORECAST is the reusable question, and it is not a danger grid"*.

Part of #367. The corking bug it reproduces is on #26.

### DIFF
--- a/docs/decisions.md
+++ b/docs/decisions.md
@@ -8543,6 +8543,93 @@ brainstorm seed, but the tool reads it, so it is part of the tool's output. Veri
 nine chapters, every `--lord-floor` table and all four curve reports (authored + three modes):
 **only ch02's numbers move.**
 
+### A rescue-fuse FORECAST is the reusable question, and it is not a danger grid (2026-09-05, #367, #26)
+
+#367 asked whether the difficulty model has a spatial element that was designed and never
+wired. The answer for a full per-tile danger grid is: not built, and deliberately out of
+scope here -- that is a different, much larger tool. What chapters actually need, repeatedly,
+is narrower: *for a protected tile (a rescue boat, a defend objective, a fragile NPC), which
+enemies can get a firing position on it, on what turn, for how much damage per phase, and
+therefore when does it die?* `tools/rescue_forecast.py` answers exactly that, reusing (never
+reinventing) three things that already existed on three different desks:
+
+- `map_placement_preview.foot_reach` -- the contested (turn-1-body-blocked) Dijkstra
+  `reached_on_contested:` already uses.
+- `difficulty`'s stat resolution (`_entry_combatants`, `_class_base`, `on_terrain`) --
+  the SAME donor-derived numbers the parity gate grades a chapter's force on.
+- `fe_combat.damage_per_round` -- the decomp's own combat math, never a second model.
+
+**The five pieces.** (1) `firing_cells(terrain, target, weapon_range)`: every cell at
+Manhattan distance 1..range that a foot unit can stand on -- FE8 has no line of sight, so
+this is pure distance, not a walk (`decisions.md` -> "What terrain cannot do is stop a
+ranged weapon"). (2) `arrival_to_cells`: the turn an enemy first stands on ANY of those
+cells, walking the CONTESTED map -- `None` (never engages) is a first-class result, not an
+error. (3) damage/phase is `fe_combat.damage_per_round` with the target resolved on its
+real terrain. (4) `sink_band`: HP / damage-per-phase is a MEAN, not a fact (`decisions.md`
+-> "A rescue clock is a HIT RATE"), so the sink turn is reported as a band -- modeled as the
+first-passage time of a random walk with the phase's own mean and variance (the Wald /
+inverse-Gaussian approximation: `E[phases] = hp/mu`, `Var[phases] = hp*sigma2/mu**3`), a
+heuristic width stated as exactly that in the docstring, not a claimed percentile. (5)
+`concurrent_attacker_cap`: when several enemies can reach one target, how many can attack
+it in the SAME phase is a bipartite MATCHING between attackers and the firing cells each
+one's own range reaches, not a single number split between them.
+
+**Validated against ch06's own measured numbers, not synthetic stand-ins.** boat-east's
+javelin-range firing cells are exactly `(15,12) (17,10) (17,13) (17,14)`, one melee cell
+`(17,13)`; boat-west's are `(2,17) (4,18) (4,19)`, one melee cell `(4,18)` -- both match
+their declared `door`. `merfolk-thrower` (soldier L7, javelin) computes to 2.76 dmg/phase
+against the hull's real fighting stats (19 HP / 6 Def in FOREST cover / +20 avoid) and
+**cannot reach any east firing cell on the contested snapshot -- its own allies cork all
+four.** `ice-crab` (bael L1, venin claw) computes to 2.82 dmg/phase, reaches its door on
+turn 2, forecast sink band ~turn 6-12 (expected 9) against a declared 8 -- inside the band.
+
+**The finding this reproduces, in one number: ch06's declared "sinks the hull on turn 7"
+(east) describes a unit that never arrives.** That prose lives in `difficulty_note:`, not a
+schema field -- nothing was asserting it, so nothing broke when it stopped being true. This
+is the SAME shape as the already-confirmed #26 bug (`decisions.md` -> "AI_B is the APPROACH
+and AI_A is the ACTION"): the tooling that measures a chapter's clock did not exist, so the
+design note was never checked against the map it now sits on.
+
+**The guard is ADVISORY, not a gate, on purpose.** `check.check_rescue_fuse_forecast` prints
+(never appends to `fail`) when a declared `rescue_pursuers:` id cannot reach a firing cell
+for any of its chapter's `rescue_boats`, and would check a declared numeric fuse against the
+forecast band if a chapter ever adopts an (currently unused, forward-looking) `declared_fuse:`
+field. Flipping the reachability half to a hard gate is a follow-up left for Nicolas once
+ch06's east pursuer is resettled (move the pursuer, or re-declare which hull is its clock) --
+this PR does not touch ch06's YAML to make the guard pass, because the finding is real and
+the fix is a design call, not a bug in the tool. `chapter_status.loose_ends` surfaces the
+identical finding for `make chapter CH=ch06`, sharing `check._fuse_forecast_findings` so the
+two never drift into two different opinions about the same chapter.
+
+**Same code path, two latent bugs fixed alongside it (map_placement_preview.py, not
+rescue_forecast.py's own code).** `units_reaching` iterated `chapter['enemy_units']` only,
+though its docstring claimed reinforcements were included -- true only because ch06 happens
+to keep its Difficult-only wave inside `enemy_units`. It now reads
+`build_campaign.chapter_roster_entries`, every roster key. `enemy_bodies` never read past
+`enemy_units:` at all, so a `reinforcements:`/`enemy_reinforcements:` entry's body was
+invisible to the contested Dijkstra's blockers regardless of timing -- and a naive fix that
+widened the loop without ALSO fixing the timing test would have flipped the bug rather than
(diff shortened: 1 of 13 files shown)


NicoviGH:
## `/code-review high`, four angles — two fixed, rest noted for follow-up

Four independent review agents found six candidate issues. Verified each by reproducing it, fixed the two that were real correctness/crash risks (three of the four angles independently converged on the first one), and I'm leaving the rest as follow-up rather than expanding this PR further.

### Fixed

**1. `_fuse_forecast_findings` could `TypeError` on the exact input its own contract exists to survive.** `pursuer_forecast` returns a row with `arrival_turn` set but every sink field `None` when the attacker reaches a firing cell yet deals zero true damage there (`sink_band`'s own `None` case — 0 hit chance, or an effectiveness mismatch). The declared-fuse loop compared `r.sink_low <= declared <= r.sink_high` over every such row unfiltered. Unreached today (no chapter has adopted `declared_fuse:` yet), but the day one does and its only reaching pursuer can't scratch it, this crashes — inside the function whose docstring and dedicated test (`test_the_check_never_fails_the_build`) promise it never fails the build. Fixed by also requiring `sink_low is not None`.

**2. `check_rescue_targets` had no guard around `units_reaching`, which this PR just widened to call `difficulty.enemy_ai_bytes` on every roster key.** That raises on an entry with neither `donor:` nor `ai_override:` — a normal mid-draft state. `check.py`'s `main()` runs every check with zero exception isolation between them, so one ungrounded `reinforcements:` entry on a future rescue chapter would crash the *entire* `check.py`, every other gate along with it. Fixed with the same try/except-and-skip idiom the function already uses two lines above for an unbuilt map.

Both reproduced with a failing test before the fix — #2 via a `check._chapters` monkeypatch (first precedent in that test file, matching the pattern already used in `chapter_status.py`'s tests).

```
tools/test_check_rescue_fuse_forecast.py: 14 OK
tools/test_check_rescue_targets.py: 9 OK
make check: green, drift clean
```

### Left as follow-up (all latent, none touch a gate or a live chapter's numbers)

- `rescue_forecast.target_combatant` bypasses `difficulty._one_enemy`'s autolevel/class-tag/personal-line handling. A no-op today (ch06's boats are level-1, no-personal `CLASS_FLEET`), silently wrong the day a boat entry adopts a level, a class tag, or a personal line.
- `firing_cells` / `arrival_to_cells` / `_boat_terrain_name` each re-derive logic that already exists inlined elsewhere (`units_reaching`'s Manhattan-range check, `reached_on`'s Dijkstra-to-arrival-turn glue, `difficulty.vanilla_terrain_at`) rather than sharing one implementation. If the underlying rule ever changes, both copies need the same edit by hand.
- `chapter_status.py` grew a third copy of the `try: import X except ImportError: None` template (now three: `_preview_module`, `_build_campaign`, `_rescue_forecast_module`).
- `map_placement_preview.placed_units` — the tool's PNG *render* path, not any gate-consumed path — still reads `enemy_units` alone, so a `reinforcements:`-key wave renders invisibly on a concept PNG even though every gate now accounts for it correctly.

Happy to open a follow-up issue for these if you want them tracked rather than left as this comment.

NicoviGH:
## Correction: the four reuse findings are fixed too

My earlier comment left these as follow-up on the grounds that they were latent. Nicolas pushed back, and he was right for most of them — that reasoning doesn't survive contact with *where the code lives*. These aren't pre-existing debt this PR walked past; they're code this PR itself wrote, or the third instance of a bug in a file whose other two instances it already claims to have fixed. A change that says "same code path, two latent bugs fixed alongside it" and leaves the third copy of that bug in that same file has fixed two thirds of a pattern.

**Fixed in `de39e0e`** (code) and **`dbecc71`** (the ADR, which still carried the stale "left as follow-up" paragraph):

1. **`placed_units` was that third copy.** `enemy_units` alone, plus a `late` flag testing `arrives_turn` alone — which reads a `trigger_turn` wave as turn-1, exactly backwards. Now iterates `ENEMY_ROSTER_KEYS` and asks `entry_is_turn1`, OR'd with `hard_mode_only` (a MODE gate — a different axis `entry_is_turn1` deliberately doesn't model, so the two conditions stay OR'd rather than one replacing the other).
2. **`target_combatant` bypassed `difficulty._one_enemy`**, so it skipped autolevel, `CLASS_TAGS` (which `fe_combat` reads to resolve effective weapons) and any personal line. No-op while ch06's boats are level-1 no-personal `CLASS_FLEET`; silently wrong the day a `rescue_boats:` entry declares otherwise.
3. **`firing_cells` moved to `map_placement_preview`** — it's a terrain/range primitive, same family as `foot_reach`, and it was a second copy of the Manhattan check `units_reaching` already ran inline. `units_reaching` now calls it; `rescue_forecast` aliases it.
4. **`reached_on` and `arrival_to_cells` were mirror-image glue** over the same two primitives, differing only in which end was plural. `arrival_turn_to` takes both ends as lists and serves either direction.
5. **`chapter_status`'s third `try/import` copy** → extracted `_try_import`; the three wrappers stay separately named because tests monkeypatch them individually.

## One I refused, and I want this on the record

The review also called `_boat_terrain_name` a duplicate of `difficulty.vanilla_terrain_at`. **It isn't, and unifying them would have been a correctness regression.** `vanilla_terrain_at` resolves a tile on a *vanilla donor layout by name*; `_boat_terrain_name` reads *our own compiled grid*, already in hand. Those differ by construction — `decisions.md` → *"ch06 departs from its donor's terrain in 21 declared cells"* — and **both boat tiles are among those 21** (`TILE_2E`/`VILLAGE_REGULAR` → `FOREST` via `terrain_divergence`). Routing the boat read through the donor lookup would swap the hull's cover bonus for the donor's terrain and move every sink band in the forecast.

A shared *shape* is not a shared *question*. I verified this against the actual data rather than taking the finding at face value.

## Verification

Output diff across all nine chapters over `placed_units`, `reached_on` (contested **and** uncontested), `units_reaching`, and the full `chapter_forecast` (arrival, damage/phase, sink band):

```
8c8
<   placed_units n=5 late=0 [... 5 bodies ...]
---
>   placed_units n=7 late=2 [... + ('rear-raiders', (0, 6), True), ('rear-raiders', (0, 7), True)]
```

That single ch02 hunk is the entire diff. It's simultaneously the proof that findings 2–5 are behaviour-preserving refactors and that finding 1 was a real bug.

151 tests across the five touched suites, each new test written first and watched fail. `make check` green, drift clean.

**Unrelated observation while diffing:** `terrain_grid` raises `KeyError: 'tileset'` for ch00/ch01/ch02 — their map JSONs predate the `tileset` key that ch03+ carry. Pre-existing, untouched by this PR, and nothing currently needs terrain for those chapters. Flagging rather than fixing; say the word if you want it picked up.

